PR #288, opened 2015-03-30
Fix two bugs

johan12345:
This fixes the following (quite critical) bugs:
- NoSuchMethodError on API < 17 for `RelativeLayout.LayoutParams.removeRule()` (strangely, Android Studio didn't warn me about this). This prevented opening SearchResultDetailFragment on those devices.
- ClassCastException for all the StepTasks (Prolong, Cancel, Prolong All, etc.). This prevented running those tasks. I currently cannot test if it works now, could you test it by prolonging something?
Maybe we should do a quick bugfix update.


raphaelm:
Prolonging doesn't work, I sent you the crash report via e-mail.
Can't you test with starting a reservation process which you do not complete?

johan12345:
Ah, good idea. I'll try.

johan12345:
I fixed it, see #290

PR #297, opened 2015-04-05
Manage dependencies via gradle

### LINKED ISSUE #296: Pull remaining jar dependencies from jCenter/mavenCentral
There are quite some jar files at https://github.com/raphaelm/opacclient/tree/master/opacclient/meaningdetector/libs , most of them should be available from the default repositories.

krt16s:
Please review and test my changes since I don't do much actual Android development. This closes #296 .


raphaelm:
Thank you very much for the feedback and the pull request!

I'll wait for a comment from @johan12345 whether he had a reason to include those JARs directly or whether he just forgot them in the transition of ``meaningdetector/`` to gradle.

johan12345:
I just forgot them during the transition. But we should investigate again if all those dependencies are needed, since they are probably already referenced through the dependency on the `test` module.

Issue #281, opened 2015-03-15
Code style refurbishment


raphaelm:
It's well over time that we clean up our codebase. Android Studio can help us, but we should do it at a time with no open branches/pull requests so I guess it will be a good time soon.

I recommend that we use https://source.android.com/source/code-style.html as a base. I see the following tasks should be done:

- [x] Convert all tabbed indentation to spaces, re-indent everything
- [x] Re-order class members(?)
- [x] Put simple block expressions (``if`` without ``{…}``) on *one* line with their body
- [x] Enforce line length of 100 characters
- [x] Order all import statements
- [ ] Re-Think all our (empty or to generic) exception handlers
- [x] Ship a Android Studio style configuration with the project
- [x] Remove duplicate code by consolidating ``catch``-Statements
- [ ] Enforce this rules for the future using git hooks / a CI server / …
- [x] Remove german acronyms (e.g. "tvEntlHeader")
- [x] Make view IDs meaningful (avoid e.g. "textView1")
- [ ] Add more Javadocs
- [x] Get rid of ``Object...`` attributes (see comments below).
- [x] Reduce number of findViewById() calls by saving found Views in variables to improve performance and readability.

cc @johan12345

johan12345:
More ideas:
[] Remove german acronyms (e.g. "tvEntlHeader")
[] Make view IDs meaningful (avoid e.g. "textView1")
[] Add more Javadocs (?)

johan12345:
> Rename private fields (?)
Do you mean [This section](https://source.android.com/source/code-style.html#follow-field-naming-conventions), which suggests to name private fields starting with "m" and static fields with "s"? I didn't do that before and it's not in the ["official" naming conventions](http://www.oracle.com/technetwork/java/codeconventions-135099.html), but if you want to, we can set that in Android Studio's code style settings, too.
> Enforce this rules for the future using git hooks / a CI server / ...
Is there some software/service you have in mind? After a short search I didn't find anything like that compatible with Java.

Should we also use Java 7 features (compatible with Android since the [KitKat Build Tools](http://tools.android.com/tech-docs/new-build-system/user-guide#TOC-Using-sourceCompatibility-1.7), but also with earlier versions) to simplify our code, for example by collapsing things like
```
try {
...
                } catch (java.net.SocketException e) {
                    exception = e;
                } catch (NoHttpResponseException e) {
                    exception = e;
                } catch (OpacErrorException e) {
                    exception = e;
[...]

PR #300, opened 2015-04-10
Even more Code style improvements (#281)

johan12345:


raphaelm:
``JsonKeyIterator.java:14: error: JsonKeyIterator is not abstract and does not override abstract method remove() in Iterator`` :)

johan12345:
Oh, I'll fix that later.

johan12345:
Now it's fixed :)

PR #313, opened 2015-06-14
App Install Location

SweetLion:
The internal memory of my device is quite small, this PR should allow users to easily move the app to the microSD.

"To allow the system to install your application on the external storage, modify your manifest file to include the android:installLocation attribute in the element, with a value of either "preferExternal" or "auto"." (source: https://developer.android.com/guide/topics/data/install-location.html)

Thank you very much for your awesome app :)

### DIFF
--- a/opacclient/opacapp/src/main/AndroidManifest.xml
+++ b/opacclient/opacapp/src/main/AndroidManifest.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    android:installLocation="auto"
     package="de.geeksfactory.opacclient">
 
     <uses-permission android:name="android.permission.INTERNET" />


johan12345:
Sorry, but if we allow the app to be moved to the SD card, this will break notifications because we will no longer receive the ACTION_BOOT_COMPLETED broadcast.

raphaelm:
Hello SweetLion,

thank you for the suggestion! As Johan pointed out, this is not as easy as it looks.

Johan, do you think it would suffice to handle ``ACTION_EXTERNAL_APPLICATIONS_AVAILABLE`` the same way as ``ACTION_BOOT_COMPLETED``?

johan12345:
@raphaelm That could work, but I don't have a device with the ability to move apps to the SD card to test that.

raphaelm:
Me neither.

SweetLion:
I could test if it works on a microSD if you could send me the APK.

The only problem I see is what happens when I plug the smartphone into the PC because then the microSD is not available for the app (data connection). Other apps using this feature just close and disappear from the app drawer and reappear after disconnecting from the PC. Using the app while charging via wall charger or in normal use should work.
Perhaps we find some others to check it because I fear that it behaves different on other devices.

johan12345:
In my [sdcard branch](https://github.com/johan12345/opacclient/tree/sdcard) I have applied your changes and added:
- handling ACTION_EXTERNAL_APPLICATIONS_AVAILABLE to re-schedule alarms for the notification service
- notification for debugging, see below

I uploaded the APK [here](https://www.dropbox.com/s/zniyhg3h78l8p81/opacapp-debug.apk?dl=1), so you can test it with your microSD capable device. We also found another person who can test it.
Here are the steps for testing:

1. Install the APK on your device and move it to the SD card. You do not need to uninstall the Play Store version.
2. The app will appear as a separate icon in the launcher as "Web Opac (debug)". You do not need to set up an account, but you can do so if you want.
3. Reboot your device
4. A few minutes  (on my device it took up to 10 minutes) after the reboot is completed, a notification with the text "OpacClient background service is working" should appear. If you swipe it away, it should come again in roughly a minute.
5. Swipe the notification away and remove the SD card -> It should not appear again
6. Put the SD card in again -> The notification should appear again after a few minutes
7. Uninstall the debug version on the app, otherwise it will use the battery heavily because it is set to run the notification service every minute

SweetLion:
Hi @johan12345, I could move the app to the microSD but I do not get the notifications "OpacClient background service is working".

SweetLion:
The app runs very fine on the microSD. I cannot see any differences.

johan12345:
The second person who tested it also said that the notification did not appear, but the battery was used quite heavily by the app, so the background service probably ran every minute as intended.

@raphaelm Do you have an idea why the notification doesn't show up? I added it [here](https://github.com/johan12345/opacclient/commit/66b8d873f4c0f64c6d1e52fc7a5c9f9a7fd6891b) and [here](https://github.com/johan12345/opacclient/commit/2b8a9a087f2ca0ff216184f401e3081674417354). Should we do further testing or should I remove the debugging code from my sdcard branch and send a pull request?
@SweetLion Did you get normal notifications from the app when a lent book is expiring, if you have that option activated in the settings? That is the only feature that will obviously not work if the background service is not running.

SweetLion:
@johan12345, I did not use the option with the lent book.

raphaelm:
No, I don't see the reason :(

raphaelm:
@SweetLion Would you try enabling this option? Also try setting up an account with username and password?

SweetLion:
I will do so.

SweetLion:
Unfortunately the notifications do **not** work properly when the app is moved to the microSD.

raphaelm:
I'm sorry to hear that, but thank you very much anyway for proposing the idea and testing.

johan12345:
The reason why this didn't work is probably this issue which has been open on the Android issue tracker since 2010:
https://code.google.com/p/android/issues/detail?id=8485

PR #324, opened 2015-07-22
Support library 22.2

johan12345:
I updated to the new support library and implemented the following thins using the new features of the Design Library:

- Floating Action Button (3rd party library not needed anymore) -> The icons on the button got a little smaller, don't know if we should restore that
- Collapsing toolbar in SearchResultDetailFragment (nicer animation, but no support for writing the title in 2 lines anymore)


raphaelm:
Would you mind updating the .travis.yml for the new build tools version? :)

raphaelm:
I suspect the SDK package is named ``build-tools-23.0.0-preview``

johan12345:
It doesn't seem to be named like this. Maybe it is not available on Travis yet?

raphaelm:
Uhm. Do we need them already?

johan12345:
I don't remember if I needed to use it because it didn't compile otherwise
or because Android Studio 1.3 prompted me to update it.

johan12345:
The old version seems to work too, so I downgraded it now.

PR #337, opened 2015-09-09
Add implementation for web-opac.at API

simon04:
This adds a first implementation plus definition of three libraries in Tyrol, Austria.

I'm happy to hear hints for improvements. :smile:

### DIFF
--- a/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/BaseApi.java
+++ b/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/BaseApi.java
@@ -33,6 +33,7 @@ import org.apache.http.client.entity.UrlEncodedFormEntity;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.client.protocol.ClientContext;
+import org.apache.http.client.utils.URIBuilder;
 import org.apache.http.client.utils.URLEncodedUtils;
 import org.apache.http.message.BasicNameValuePair;
 import org.apache.http.protocol.BasicHttpContext;
@@ -45,8 +46,8 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.InterruptedIOException;
 import java.io.UnsupportedEncodingException;
+import java.net.URISyntaxException;
 import java.net.URLDecoder;
-import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -472,27 +473,12 @@ public abstract class BaseApi implements OpacApi {
         this.stringProvider = stringProvider;
     }
 
-    protected String buildHttpGetParams(Map<String, String> params,
-                                        String encoding) throws UnsupportedEncodingException {
-        String string = "?";
-        for (Map.Entry<String, String> pair : params.entrySet()) {
-            String name = URLEncoder.encode(pair.getKey(), encoding);
-            String value = URLEncoder.encode(pair.getValue(), encoding);
-            string += name + "=" + value + "&";
+    public static String buildHttpGetParams(List<NameValuePair> params)
+            throws UnsupportedEncodingException {
+        try {
+            return new URIBuilder().addParameters(params).build().toString();
+        } catch (URISyntaxException e) {
+            throw new RuntimeException(e);
         }
-        string = string.substring(0, string.length() - 1);
-        return string;
-    }
-
-    protected String buildHttpGetParams(List<NameValuePair> params,
-            String encoding) throws UnsupportedEncodingException {
-        String string = "?";
-        for (NameValuePair pair : params) {
-            String name = URLEncoder.encode(pair.getName(), encoding);
-            String value = URLEncoder.encode(pair.getValue(), encoding);
-            string += name + "=" + value + "&";
-        }
-        string = string.substring(0, string.length() - 1);
-        return string;
     }
 }
--- a/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/BaseApi.java
+++ b/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/BaseApi.java
@@ -472,6 +472,17 @@ public abstract class BaseApi implements OpacApi {
         this.stringProvider = stringProvider;
     }
 
+    protected String buildHttpGetParams(Map<String, String> params,
+                                        String encoding) throws UnsupportedEncodingException {
+        String string = "?";
+        for (Map.Entry<String, String> pair : params.entrySet()) {
+            String name = URLEncoder.encode(pair.getKey(), encoding);
+            String value = URLEncoder.encode(pair.getValue(), encoding);
+            string += name + "=" + value + "&";
+        }
+        string = string.substring(0, string.length() - 1);
+        return string;
+    }
 
     protected String buildHttpGetParams(List<NameValuePair> params,
             String encoding) throws UnsupportedEncodingException {
--- a/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/ReadOnlyApi.java
+++ b/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/ReadOnlyApi.java
@@ -1,0 +1,67 @@
+package de.geeksfactory.opacclient.apis;
+
+import org.json.JSONException;
+
+import java.io.IOException;
+
+import de.geeksfactory.opacclient.objects.Account;
+import de.geeksfactory.opacclient.objects.AccountData;
+import de.geeksfactory.opacclient.objects.DetailledItem;
+import de.geeksfactory.opacclient.objects.Library;
+
+/**
+ * For "read-only" API implementations, i.e. without account support, some abstract method
+ * implementations are provided.
+ */
+public abstract class ReadOnlyApi extends BaseApi {
+
+    @Override
+    public boolean isAccountSupported(Library library) {
+        return false;
+    }
+
+    @Override
+    public boolean isAccountExtendable() {
+        return false;
+    }
+
+    @Override
+    public String getAccountExtendableInfo(Account account) throws IOException {
+        return null;
+    }
+
+    @Override
+    public ReservationResult reservation(DetailledItem item, Account account,
+                                         int useraction, String selection) throws IOException {
+        return null;
+    }
+
+    @Override
+    public ProlongResult prolong(String media, Account account, int useraction,
+                                 String selection) throws IOException {
+        return null;
+    }
+
+    @Override
+    public ProlongAllResult prolongAll(Account account, int useraction, String selection)
+            throws IOException {
+        return null;
+    }
+
+    @Override
+    public CancelResult cancel(String media, Account account, int useraction,
+                               String selection) throws IOException, OpacErrorException {
+        return null;
+    }
+
+    @Override
+    public AccountData account(Account account)
+            throws IOException, JSONException, OpacErrorException {
+        return null;
+    }
+
+    @Override
+    public void checkAccountData(Account account)
+            throws IOException, JSONException, OpacErrorException {
+    }
+}
--- a/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/WebOpacAt.java
+++ b/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/WebOpacAt.java
@@ -21,6 +21,7 @@
  */
 package de.geeksfactory.opacclient.apis;
 
+import de.geeksfactory.opacclient.searchfields.DropdownSearchField;
 import org.apache.http.client.utils.URIBuilder;
 import org.json.JSONException;
 import org.json.JSONObject;
@@ -246,6 +247,7 @@ public class WebOpacAt extends SearchOnlyApi {
         final List<SearchField> fields = new ArrayList<>();
         addSimpleSearchField(fields);
         addAdvancedSearchFields(fields);
+        addSortingSearchFields(fields);
         return fields;
     }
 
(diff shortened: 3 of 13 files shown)


opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/ReadOnlyApi.java
```diff
@@ -0,0 +1,67 @@
...
+import de.geeksfactory.opacclient.objects.Account;
+import de.geeksfactory.opacclient.objects.AccountData;
+import de.geeksfactory.opacclient.objects.DetailledItem;
+import de.geeksfactory.opacclient.objects.Library;
+
+/**
+ * For "read-only" API implementations, i.e. without account support, some abstract method
+ * implementations are provided.
+ */
+public abstract class ReadOnlyApi extends BaseApi {
```
raphaelm: I find the idea of ``ReadOnlyApi`` interesting, it really does remove duplicate code from multiple of our APIs.
@johan12345, what do you think of it?

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/BaseApi.java
```diff
@@ -472,6 +472,17 @@ public void setStringProvider(StringProvider stringProvider) {
         this.stringProvider = stringProvider;
     }
 
+    protected String buildHttpGetParams(Map<String, String> params,
+                                        String encoding) throws UnsupportedEncodingException {
```
raphaelm: I'm not a fan of having this logic twice, once for ``Map`` and once for ``List<NameValuePair>``. I prefer ``List<NameValuePair>``, as it more accurately maps to HTTP querystring parameters – and is the same type that the Apache HTTP client expects for POST request data.

raphaelm: Still, I see that using a ``Map`` in your API implementation is a lot easier than the List of pairs (because you need to replace values multiple times).
I think it is okay like this.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/WebOpacAt.java
```diff
@@ -0,0 +1,290 @@
...
+        final Element availabilityTable = doc.select(".bibliothek table").first();
+
+        final DetailledItem result = new DetailledItem();
+        final HashMap<String, String> copy = new HashMap<>();
+        result.addCopy(copy);
+        result.setId(id);
+        result.setCover(getCover(doc));
+        result.setTitle(detailData.select("h3").first().text());
+        result.setMediaType(MEDIA_TYPES.get(getCellContent(detailTable, "Medienart|Type of media")));
+        copy.put(DetailledItem.KEY_COPY_STATUS, getCellContent(availabilityTable, "Verfügbar|Available"));
```
raphaelm: You are hardcoding this to English/German here, while other parts of the source hint at a Turkish version and there might be even more languages in other libraries. I get that there is no other way then these regexes to detect media type and reservation count, but we might take a different approach for the status field and search for the status image embedded there. What do you think, is it worth the effort?

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/WebOpacAt.java
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    @Override
+    public String getShareUrl(String id, String title) {
+        return null;
+    }
+
+    @Override
+    public int getSupportFlags() {
+        return 0;
```
raphaelm: It should be possible to add ``SUPPORT_FLAG_ENDLESS_SCROLLING`` here, which makes the UI a lot faster.

raphaelm:
I really like your implementation, it is very well-structured. I added a few minor remarks, but otherwise I would wait for a small comment from @johan12345 and then merge it. Thanks a lot!

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/ReadOnlyApi.java
johan12345: Yes, nice idea. I am not 100% satisfied with the name, however: In the literal sense, a `ReadOnlyApi` would for example also provide the `account` function, because it only *reads* data from the account. Maybe `SearchOnlyApi` or `NoAccountApi` would be better?

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/BaseApi.java
johan12345: I see only one place where a value is replaced (in `searchGetPage`). It could also be implemented using `List<NameValuePair>` by rebuilding the parameters for `searchGetPage` from the `lastQuery` (adding a `page` parameter to `buildSearchParams`). Of course, this would slow down the `searchGetPage` function, but it is probably negligible compared to the HTTP request duration. And it would have the advantage of having one single function that generates the parameters for both `search` and `searchGetPage`.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/WebOpacAt.java
```diff
@@ -0,0 +1,290 @@
...
+        put("Hörbuch", SearchResult.MediaType.AUDIOBOOK);
+        put("eMedium", SearchResult.MediaType.EBOOK);
+        // en
+        put("Book", SearchResult.MediaType.BOOK);
+        put("Periodical", SearchResult.MediaType.MAGAZINE);
+        put("CD ROM", SearchResult.MediaType.CD_SOFTWARE);
+        put("DVD", SearchResult.MediaType.DVD);
+        put("Audiobook", SearchResult.MediaType.AUDIOBOOK);
+        put("eMedium", SearchResult.MediaType.EBOOK);
+    }};
```
johan12345: Is there any way of detecting the media types independent from the language? In most other APIs we use the file names of media type icons that are shown in each search result or CSS classes.
If that's not possible, @raphaelm, what would you think about an API independent thing for detecting media types from their names, similar to the `MeaningDetector` for search fields? This would also be helpful for the `Open` API because some libraries do not have media type icons.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/WebOpacAt.java
```diff
@@ -0,0 +1,290 @@
...
+
+    @Override
+    public SearchRequestResult filterResults(Filter filter, Filter.Option option)
+            throws IOException, OpacErrorException {
+        return null;
+    }
+
+    @Override
+    public DetailledItem getResult(int position) throws IOException, OpacErrorException {
+        return null;
```
johan12345: Maybe adding a comment here that every result will have an ID would be nice

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/WebOpacAt.java
```diff
@@ -0,0 +1,290 @@
...
+        final Elements options = doc.select("select#adv_search_crit_0").first().select("option");
+        for (final Element option : options) {
+            final TextSearchField field = new TextSearchField();
+            field.setDisplayName(option.text());
+            field.setId(option.val());
+            field.setHint("");
+            field.setData(new JSONObject());
+            field.getData().put("meaning", field.getId());
+            fields.add(field);
+        }
```
johan12345: What about the "sortiert nach" (only one sorting level would be enough, I think), "Suche in" (we would need a have a SearchField type for a group of checkboxes for that or implement it as a dropdown with an option for "all") and "Alle Titel" / "nur Antolin Titel" search fields?

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/WebOpacAt.java
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    @Override
+    protected String getDefaultEncoding() {
+        return "utf-8";
+    }
+
+    @Override
+    public String getShareUrl(String id, String title) {
+        return null;
```
johan12345: You could probably use the same URL as in `getResultById` here.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/ReadOnlyApi.java
raphaelm: ``SearchOnlyApi`` sounds good to me.
How do you feel about declaring the overridden methods as ``final``, such that we do not end up with every API that originally based on ``SearchOnlyApi`` continues to do so?

johan12345: Yes, that is probably a good idea.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/WebOpacAt.java
raphaelm: I dont think there is a nother way to detect the media type here.

> If that's not possible, @raphaelm, what would you think about an API independent thing for detecting media types from their names, similar to the MeaningDetector for search fields? This would also be helpful for the Open API because some libraries do not have media type icons.

I feel that this is not necessarry *at the moment*. web-opac.at ist barely used outside Austria, OPEN not outside Germany, so the installed languages do not differ greatly. Additionally, implementing something like ``MeaningDetector`` for types is way harder, because there is no standard way of getting a *list* of media types, so we'd have to "guess" around with a lot of search queries. And it would also still need one of us to do the mapping to media types.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/WebOpacAt.java
raphaelm: Yes, I missed that. You should definitively implement this method! Just return ``opac_url + "/search?view=detail&id=" + id``

johan12345:
There is a list of more (if not all) libraries using web-opac.at [here](https://littera.eu/littera_software/referenzen.html), we could add all of them later using the `tools/add_library.py` Python script.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/BaseApi.java
simon04: Good idea, I changed that in 65f8dce11d45bdf6341e77240860f932fc42de74 and construct the query using Apache's `URIBuilder`.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/ReadOnlyApi.java
simon04: Done in 794598e18ba783d96b21520585d6bb702926d73a.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/WebOpacAt.java
simon04: I agree that it's somewhat of a hack. ASAIK, only the languages `de`, `en`, `tr` are implemented at the moment. For "Verfügbar|Available", we could make use of the status image. But especially for "Medienart|Type of media" there are no indications. The only different approach would be via the search fields: "Medienart" is associated the key `ma`. This key could be used to determine the lookup term "Medienart" …Rather complex instead of specifying two/three translations here. :smile:

raphaelm: Yes, I agree.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/WebOpacAt.java
simon04: So the `SearchField`s are not only used to specify a search criterion/filter, but also for sorting? Would that just be an indistinctive `SearchField` that can be added to the query?

"Antolin" should be rather easy to add, although I'm not overly interested in that criterion :smirk:. "Suche in/Partnerbibliothek" adds a state to performing queries, since that is set via a different HTTP request.

Another enhancement would be to obtain all possible values for certain fields (such as "Systematik|Classification", via http://www.aktirol.web-opac.at/search/adv_ac?crit=og) and add a `DropdownSearchField` instead …

simon04:
Thank you a lot for your extensive review. I addressed most aspects in the commits 794598e, cd4af0c.

> There is a list of more (if not all) libraries using web-opac.at

There is an even longer list https://www.bvoe.at/oeffentliche_bibliotheken/adressen (needs to be filtered for WebOPAC entries). I'm happy to prepare the JSON files once the implementation is set :smile:

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/WebOpacAt.java
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    @Override
+    public Set<String> getSupportedLanguages() throws IOException {
+        return null;
+    }
+
+    @Override
+    public void setLanguage(String language) {
+    }
```
johan12345: You should implement `getSupportedLanguages` and `setLanguage` so that the app can set the language depending on the device language.
`getSupportedLanguages` could do a network operation to find out the available languages, but if it is German, English and Turkish in all the libraries, you could return those codes directly as well.
The current language of the device is passed to `setLanguage`, which you should save and then use for the search and detail requests. If the language passed to `setLanguage` is not one of the supported ones, you should fall back to English (and, if that should also not be supported, German).
See also the [Javadoc](https://github.com/raphaelm/opacclient/blob/master/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/OpacApi.java#L587) and the implementation in [Pica](https://github.com/raphaelm/opacclient/blob/master/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Pica.java#L865)

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/WebOpacAt.java
johan12345: Yes, `SearchFields` for sorting are not different from the other ones (except normally being a `DropdownSearchField` instead of a `TextSeachField`). The API implementations are more or less flexible in recognizing different search fields provided by different libraries and we have the `MeaningDetector` for detecting their meanings by their name (using our [rather large JSON file](https://github.com/raphaelm/opacclient/blob/master/opacclient/opacapp/src/main/assets/meanings/general.json)) and sorting them in the form. This would also ensure that the "Antolin" field is only shown when you click on "Mehr Suchfelder anzeigen" in the app.

Obtaining all the possible values for certain fields would be quite nice, we could whitelist the ones where it makes sense based on their ID.

simon04: Sorting search fields added in d74f314.

simon04: Dropdown values are fetched for certain fields since 4b65296.

simon04:
Is anything left to be done?

raphaelm:
Nope, seems good to me!

PR #338, opened 2015-09-10
Introduce DropdownSearchField.Option

simon04:
This allows a more comprehensible and compact usage of the
DropdownSearchField. In addition, due to replacing many HashMap
instances by the simpler DropdownSearchField.Option object, the memory
footprint will be reduced.

After applying #337, the class `WebOpacAt` needs to be adapted …


raphaelm:
Seems legit to me ;)

PR #340, opened 2015-09-14
Add many libraries using web-opac.at

simon04:
The collection has been assembled from
https://www.bvoe.at/oeffentliche_bibliotheken/adressen_liste

### DIFF
--- a/opacclient/opacapp/src/main/assets/bibs/Mils.json
+++ b/opacclient/opacapp/src/main/assets/bibs/Mils.json
@@ -1,0 +1,12 @@
+{
+    "state": "Tirol",
+    "account_supported": false,
+    "title": "Bibliothek Mils",
+    "city": "Mils",
+    "country": "\u00d6sterreich",
+    "api": "web-opac.at",
+    "geo": [47.2883084, 11.5314953],
+    "data": {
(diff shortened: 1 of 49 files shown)


raphaelm:
Thank you!

johan12345:
The `title` attribute normally does not need to include the `city` name like you did it with some of the libraries, because the app will show both next to each other everywhere.

raphaelm:
Uh, I missed that. Fixed in 08a41161c70e7a6e34096b4b4e1450d2f4d4ad91

(Also, indentation etc. has become inconsistent, I fixed that in 143d11404860aaf9a0dd49aa187a2e13f6ccda0f)

opacclient/opacapp/src/main/assets/bibs/Mils.json
```diff
@@ -0,0 +1,12 @@
...
+    "account_supported": false,
+    "title": "Bibliothek Mils",
+    "city": "Mils",
+    "country": "\u00d6sterreich",
+    "api": "web-opac.at",
+    "geo": [47.2883084, 11.5314953],
+    "data": {
+        "baseurl": "http://www.buecherei-mils.webopac.at"
+    }
+}
```
johan12345: This library is missing the `information` attribute and therefore the app does not display it

PR #341, opened 2015-09-16
Some enhancements for web-opac.at API

simon04:


opacclient/tests/src/main/java/de/geeksfactory/opacclient/apis/WebOpacAtTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Test
+    public void testParseCopyReturn() throws Exception {
+        assert Objects.equals(WebOpacAt.parseCopyReturn("1 (voraussichtl. bis 31.07.2015)"), "31.07.2015");
+
+    }
+
+    @Test
+    public void testSearchFields() throws Exception {
+        getApi().getSearchFields();
```
raphaelm: @johan12345, what do you think on these tests, especially this and the next one? Do they make sense or should we incorporate them into the general unit tests? I see why they are useful, maybe more useful than the general ones (becuase they are faster to run etc.), but I don't like introducing API-specific single-library-dependent tests for such a generic thing.

simon04: Maybe I should reduce the test to `testParseTotalResults` and `testParseCopyReturn`? The other methods were helpful while developing and debugging the implementation …

raphaelm: Yes, I think so.
We'll consider adding a test for getSearchFields to the generic unit tests.
For empty search request, the expected result is not even well-defined across the different APIs (some return all media in the library, some throw errors), so writing a generic unit test should be hard.

johan12345: We already use `getSearchFields` in the generic Unit tests and test if there is at least a search field for free search or search by title, so I think an individual test for `getSearchFields` would not be necessary.
The only thing I could imagine a generic `testEmptyQuery` test doing would be to check if either a result list is returned or an `OpacErrorException` is thrown (which can show a user-understandable error message).

I also like the idea of having a different type of unit tests that only test if the APIs work in general instead of in every single library. Those could even be run on the CI server to make sure we don't destroy anything important when adjusting things in the APIs. But I understand that we should not rely on the OPAC of a specific library being online for that, too.
So the three options I see for doing this without relying on a library OPAC would be 
1. Have a demo instance of every OPAC we support under our control for running these kinds of tests, which is probably quite unrealistic for most commercial systems and doable, but probably hard to set up for open source systems like VuFind or Koha.
2. Build a mock server for every OPAC which is run by the unit tests themselves (using Jetty?) and would behave similar to the library servers, but could return prepared, static HTML responses (e.g. the same result list for every search query). Then we could run generic unit tests for all APIs on those
3. Similar to the previous one but a little simpler: Build specific unit tests for every API that could call functions like `parse_search`, `parse_result` (BTW @raphaelm : Is there a reason why you didn't use camelCase for these names?) etc. using prepared HTML. This would be "one level higher" compared to the two specific unit tests `testParseTotalResults` and `testParseCopyReturn` here.

All of these options would even allow us to test account related features, which is something we cannot do in the generic unit tests.
The latter two options would be doable in theory, but probably a great deal of work to build (and to maintain when adding new OPACs or a different version of the same OPAC).

raphaelm: Of all the three ways you proposed, I only like the third one better than testing against a specific library. The good thing is, that we do not have to provide all this unit tests at the same time, we can add them slowly during the next years -- event one unit test is better than none.

However, this leaves the scope of this PR. So @simon04, if you remove those two tests as suggested, I'll go ahead and merge.

PS: There is no reason for not using camel case there, but at least it's not a public API. ;)

simon04:
Thank you! I removed the calls to online API in 1aff118 …

PR #343, opened 2015-09-24
Updated reminder feature

johan12345:
I completely reimplemented the reminder feature and its background service. The new procedure can be roughly described as follows:

- `ReminderCheckService` was replaced with `SyncAccountService`, which is only responsible for downloading account data and saving it to the database. It is currently set to run every hour, see the comment in `SyncAccountAlarmListener` for explanation. `SyncAccountService` uses [commonsguy/cwac-wakeful](https://github.com/commonsguy/cwac-wakeful/) to be assured stay awake while working.
- After downloading account data, lent media will be grouped by its deadline dates and corresponding entries will be saved to the new `alarms` DB table. They contain the list of items expiring and the scheduled notification time. All saved notifications from the DB will be scheduled using `AlarmManager` and rescheduled every time the device is rebooted. This allows notifications to be shown in times without internet connection, which is why the corresponding note in the settings was removed.
- Notifications will contain a list of expiring media and two buttons: `Don't remind again` and `Remind me again at...` ("Snooze"). The latter will open dialogs to select the date and time (these dialogs don't use Material Design in older Android versions, because they are not included in the Support Library).
- Notifications that were swiped away or clicked to open the app will automatically be re-scheduled between 12 and 2 hours after dismissal (depending on the time left until the media needs to be returned). This functionality can be disabled in the settings
- Known problems:
  - Notifications displaying while the app is being updated or force-closed will disappear and not come again
  - ~~The titles of the two buttons in the notification are too long to be completely visible~~ fixed in 135dd31.
  - `SyncAccountService` will also run when notifications are disabled. This is not necessary. The simplest option would be to check the settings before doing work in `SyncAccountService`, but this would still wake the device every time it runs.

In this process, I refactored the data structure for lent and reserved account items so that we have proper classes for those objects (instead of the previous `Map<String, String>` which actually was a remainder from the implementation using the `ContentValues` for Android's SQLite database directly prior to that). They also use Joda Time instead of timestamps for all dates.
Refactoring all the APIs to use Joda Time was quite complex and could cause issues because it was not tested for all the APIs, especially for the fields `readyDate` and `expirationDate` in `ReservedItem`, which only used Strings and no timestamps before.

The notification functionality and also the refactored APIs (see above) were not tested much, so we should do that before merging. Testing it with the larger amount of test accounts that @raphaelm has access to is probably a good idea.


raphaelm:
> The titles of the two buttons in the notification are too long to be completely visible

How about "Remind Later" (or even only "Later") and "Discard" as button descriptions? For German: "Später"/"Verwerfen"?

raphaelm:
> Notifications displaying while the app is being updated or force-closed will disappear and not come again

Is there a possible fix that you know of or do we just have to live with it? I don't really like it as I force-close apps more often than I restart my phone.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/objects/AccountItem.java
```diff
@@ -0,0 +1,148 @@
+package de.geeksfactory.opacclient.objects;
+
+public abstract class AccountItem {
+    private long account;
+    private String title;
+    private String author;
+    private String format;
```
raphaelm: Was there a reason for having a string here instead of moving ``SearchResult.MediaType`` to e.g. ``OpacApi`` and using it in both places? This would enable us to show icons in the acocunt view in the future. String serialiazation for the database should be done at database communication time.

opacclient/opacapp/src/main/AndroidManifest.xml
```diff
@@ -136,8 +138,11 @@
             android:finishOnTaskLaunch="true"
             android:launchMode="singleInstance"
             android:theme="@style/Theme.AppCompat.Light.Dialog" />
+        <activity
+            android:name="de.geeksfactory.opacclient.frontend.SnoozeDatePickerActivity"
+            android:theme="@style/Theme.Opacapp.Transparent" />
 
-        <service android:name="de.geeksfactory.opacclient.reminder.ReminderCheckService" >
+        <service android:name=".reminder.SyncAccountService">
```
raphaelm: I guess we always use full-qualified class paths here because it makes life simpler for the Plus edition. I'm unsure whether this still holds, but we should either test it or just ontinue using a full class path.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/AccountFragment.java
```diff
@@ -1374,6 +1226,22 @@ public void onCollapseAnimationEnd() {
         refreshage();
     }
 
+    private void setHtmlTextOrHide(String value, TextView tv) {
+        if (value != null) {
```
raphaelm: Is comparison with null good enough here or should we also check for empty strings?

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/MainPreferenceActivity.java
```diff
@@ -23,4 +24,9 @@ protected void onCreate(Bundle savedInstanceState) {
     protected PreferenceFragment getNewMainPreferenceFragment() {
         return new MainPreferenceFragment();
     }
+
+    @Override
+    public void onActivityResult(int requestCode, int resultCode, Intent data) {
```
raphaelm: What exactly is the purpose of this? ;)

johan12345:
>> The titles of the two buttons in the notification are too long to be completely visible

> How about "Remind Later" (or even only "Later") and "Discard" as button descriptions? For German: "Später"/"Verwerfen"?

Yes, sounds good. Changed in 135dd31.

>> Notifications displaying while the app is being updated or force-closed will disappear and not come again

> Is there a possible fix that you know of or do we just have to live with it? I don't really like it as I force-close apps more often than I restart my phone.

I think there is no way to detect when the app is force-closed (that would make force-closing meaningless). The other way I could think of would be to re-schedule/re-show notifications if `notified && !finished`, but `NotificationManager#getActiveNotifications()` doesn't contain the corresponding notification.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/objects/AccountItem.java
johan12345: I left it as a String because I don't know if we can detect `SearchResult.MediaType`s using things like CSS classes or icon URLs in the account views of APIs currently using this property or if some simply have a text describing the media type. If the latter is the case, we would need something to convert names of media types to their `SearchResult.MediaType` values to be able to show them in the app.

opacclient/opacapp/src/main/AndroidManifest.xml
johan12345: Yes, that's probably correct. Changed in c237f04.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/AccountFragment.java
johan12345: In the best case where all APIs set a property to `null` if it is not available, this would be good enough. But that's probably not the case in all APIs. Should I change it?

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/MainPreferenceActivity.java
johan12345: Oh, I added it when I tried to fix a problem where selecting a different warning period in the settings didn't have any effect. But the real problem was I needed to return `true` instead of `false` at `MainPreferenceFragment.java:90`, so it can be removed again. Done in 3b60160.

johan12345:
The failing Travis build because of a Lint error should be fixed with ef3fea4

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/reminder/ReminderHelper.java
```diff
@@ -0,0 +1,133 @@
...
+        for (Alarm alarm : data.getAllAlarms()) {
+            if (!arrangedIds.containsKey(alarm.deadline)) {
+                data.removeAlarm(alarm);
+            }
+        }
+
+        // Find and add/update corresponding alarms for current lent media
+        for (Map.Entry<LocalDate, List<Long>> entry : arrangedIds.entrySet()) {
+            LocalDate deadline = entry.getKey();
+            long[] media = toArray(entry.getValue());
```
raphaelm: It is not a real problem, but since were doing code-review and since we are perfectionists: I don't really like the number of ``List<Long>``/``long[]``/comma-separated string conversions.

Storing comma-separated strings in a database is a [really bad](https://en.wikipedia.org/wiki/Database_normalization) design pattern in general, but **I'm absolutely okay with it here** since the alarm database works more like a cache than like a relational database. **However**, on the Java side, wouldn't be some form of ``Set`` (e.g. ``HashSet``) the appropriate representation instead of mixing arrays and lists?

So I'd suggest keeping the database as it is but use Sets throughout this file and inside the Alarm object.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/reminder/SyncAccountAlarmListener.java
```diff
@@ -0,0 +1,33 @@
...
+
+import android.app.AlarmManager;
+import android.app.PendingIntent;
+import android.content.Context;
+
+public class SyncAccountAlarmListener implements WakefulIntentService.AlarmListener {
+    // We could use a longer interval like 3 hours, but on API < 19, only the AlarmManager
+    // .INTERVAL_... constants make the AlarmManager use inexact alarm delivery for better
+    // battery efficiency. The next larger constant available would be AlarmManager
+    // .INTERVAL_HALF_DAY.
```
raphaelm: We really should discuss this, as it greatly affects both battery life and mobile data tariffs as well as the quality of the notification service.

During normal operation, half a day *should* really be enough, shouldn't it? I've never heard of a library with lending times less than a week, so if you lend out a new item, it should be enough for all notification purposes if the app only sees it half a day later.

A half day is **not enough** if a sync attempts fail, either because of a network connection failure or because the notification service has been limited to WiFi and WiFi is not available, because "half day" is long enough to coincidentally always occur when I'm not in a WiFi network. However, in this case we could go without inexact repetition.

So if I do not miss anything, we should absolutely get away with *regular* runs every (inexact) half day and *repeated runs due to failures* at fixed intervals (e.g. 1 hour after the failure).

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/reminder/SyncAccountService.java
```diff
@@ -0,0 +1,99 @@
...
+            if (BuildConfig.DEBUG) Log.i(NAME, "Loading data for Account " + account.toString());
+
+            try {
+                Library library = app.getLibrary(account.getLibrary());
+                if (!library.isAccountSupported()) continue;
+                OpacApi api = app.getNewApi(library);
+                AccountData res = api.account(account);
+                if (res == null) continue;
+                data.storeCachedAccountData(account, res);
+                data.close();
```
raphaelm: You are using ``data.close`` after every account but ``data.open`` only once. Do I miss something or will this fail with more than one account?

Also, we should not keep the DataSource open during network operations to prevent concurrency issues. Therefore, we should close it after line 76 and open it again just before line 87.

opacclient/opacapp/src/main/res/values/strings.xml
```diff
@@ -59,19 +59,13 @@
     <string name="notification_service">Expiring items</string>
     <string name="notification_service_desc">Notify, if an item is about to expire (no warranties given)</string>
     <string name="notification_service_wifionly">Check on WiFi only</string>
-    <string name="notification_service_wifionly_desc">Does not download information over mobile networks. (Therefore, no warnings without WiFi!)</string>
+    <string name="notification_service_wifionly_desc">Does not download information over mobile networks.</string>
```
raphaelm: I guess we should keep *some* kind of warning that this setting *might* keep you from getting notifications if you're out of WiFi for a long time.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/reminder/ReminderHelper.java
johan12345: Regarding the comma-separated strings in the database: Is there even a better way? Storing JSON or binary data instead? Or doing it like a real relational database by storing the alarm IDs in the lent items table instead of the lent item IDs in the alarm table (which we already do to a certain extent, because every lent item could be linked to its corresponding alarm using its deadline date)?

Regarding the `List<Long>/long[]/Set<Long>` issue: I agree that a set would be the better data structure to represent this: The order of the lent items corresponding to an alarm is not important. But it would cause the expired items to be displayed in the notification in a different order every time the notification is shown, which *could be* confusing, and if we consequently also changed the `List<LentItem>` and `List<ReservedItem>` in the `AccountData` class to `Set`s, that would *definitely* be confusing because the order would change every time the user re-opens or refreshes the account view.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/reminder/SyncAccountAlarmListener.java
johan12345: So you suggest to set the regular interval to `AlarmManager.INTERVAL_HALF_DAY` and tell `SyncAccountService` to schedule an additional alarm one hour later if it fails?
I agree that these times would be enough for normal operation, but don't how to implement it in the best way:

If we have no internet connection (or no WiFi with the corresponding setting activated, or a library website is offline) for a longer time (say, a week), the service would run every (inexact) half day and schedule itself to run one hour later. Then, the "one hour later" runs would reschedule themselves to run one hour later and won't stop doing that when the next "every half day" run is reached. This would result in the service running ~14 times per hour at the end of the week (and probably not concentrated around one point in the hour due to the inexactness of the "every half day" alarms).

Another way to do it would be to determine the interval to the next run of the service in the service itself and let it only schedule the next run, like we did it before. The problem with this way is that if one run of the service is interrupted (due to low memory or a crash or whatever), the service probably won't run again until the next device reboot.

We could also use Android's [`JobScheduler`](https://developer.android.com/reference/android/app/job/JobScheduler.html) instead of `AlarmManager` for scheduling service runs, so it won't even run when there's no Internet connection available. The problem is it is only available on API 21 and above. There's also [`GcmNetworkManager`](https://developers.google.com/cloud-messaging/network-manager), but it only works on devices with Google Play Services.

opacclient/opacapp/src/main/res/values/strings.xml
johan12345: Hmm, so we keep the same warning? Or do you have another (short) text that would fit?

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/reminder/SyncAccountService.java
johan12345: Correct! Fixed in b2e882a.
I also found another bug related to multiple accounts and fixed it in 7721657.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/reminder/ReminderHelper.java
raphaelm: > Regarding the comma-separated strings in the database: Is there even a better way?

The "normalized" way would be to have three tables: ``alarm``, ``lent_item``, ``lent_item_on_alarm``. But I think it is overkill here as we do not need the features of relational databases.

> But it would cause the expired items to be displayed in the notification in a different order every time the notification is shown

Well, we could sort it at display time. 

> and if we consequently also changed the List<LentItem> and List<ReservedItem> in the AccountData class to Sets, that would definitely be confusing because the order would change every time the user re-opens or refreshes the account view.

You're right about that, I take my suggestion back. 

On the other hand, this "convert to array, than use Array.equals" compares order-sensitive. Should we use a Set instead of an array just here?

opacclient/opacapp/src/main/res/values/strings.xml
raphaelm: I'll think about another text.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/reminder/SyncAccountAlarmListener.java
raphaelm: > If we have no internet connection (or no WiFi with the corresponding setting activated, or a library website is offline) for a longer time (say, a week), the service would run every (inexact) half day and schedule itself to run one hour later. Then, the "one hour later" runs would reschedule themselves to run one hour later and won't stop doing that when the next "every half day" run is reached. This would result in the service running ~14 times per hour at the end of the week (and probably not concentrated around one point in the hour due to the inexactness of the "every half day" alarms).

This should, of course, not happen, but there are things we can do to prevent it. We only need a way for the scheduled run to determine if there already is an hourly run scheduled and only reschedule itself on failure if there isn't, right?

>We could also use Android's JobScheduler instead of AlarmManager for scheduling service runs, so it won't even run when there's no Internet connection available. The problem is it is only available on API 21 and above. There's also GcmNetworkManager, but it only works on devices with Google Play Services.

Sadly, this is both not an option currently. :(

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/reminder/ReminderHelper.java
johan12345: > Well, we could sort it at display time.

Yes, but by what? By ID?

> On the other hand, this "convert to array, than use Array.equals" compares order-sensitive. Should we use a Set instead of an array just here?

So then change the `long[]` in the `Alarm` class to a `List<Long>` and convert it to a `Set` for comparison at `ReminderHelper.java:68`, and use a `Map<LocalDate, Set<Long>>` for the `arrangedIds` variable?

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/reminder/SyncAccountAlarmListener.java
johan12345: > We only need a way for the scheduled run to determine if there already is an hourly run scheduled and only reschedule itself on failure if there isn't, right?

We could save the current interval (either 1 hour or half a day) in the `SharedPreferences` and let the `SyncAccountAlarmListener.scheduleAlarms` function use this to schedule alarms in this interval. There, `firstStart` would also need to be set to the current time plus the interval.

Then, every time the `SyncAccountService` runs, it will update the preference value depending on if it failed or not and reschedule the Alarms by calling `WakefulIntentService.cancelAlarms` and `WakefulIntentService.scheduleAlarms`. If the service does not get to this point for whatever reason, the alarms will stay as they were, so it will still be called again the next time.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/reminder/ReminderHelper.java
raphaelm: Hmm. Sounds as complicated as it is curerntly. There does not seem to be a good solution here, maybe we should just keep it as it is :D

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/reminder/SyncAccountAlarmListener.java
raphaelm: Sounds reasonable to me.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/reminder/ReminderHelper.java
johan12345: We could also leave the `if (!Arrays.equals(media, alarm.media)) {` completely and update the alarm every time. And then use `List` everywhere instead of converting to arrays.

johan12345:
In 76708d7, I added the ability to update the `notification_warning` value from milliseconds to days after the app was updated and with f0be0b0, the app will also call the `ReminderHelper.generateAlarms` function when account data is updated inside the app.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/reminder/SyncAccountAlarmListener.java
johan12345: Something like d89fd58?

raphaelm:
I just started testing and I'll just leave some notes here.

* Due to switching between branches, I had an ``java.io.FileNotFoundException: bibs/Rostock.json`` exception. I wondered if there is a way to catch *all* exceptions in the service and present the ACRA notifications when the app is started for the next time. *Nobody* likes an app that loudly crashes while not being used.

Testing todo:

- [ ] Test over a longer period with WiFI
- [ ] Test over a longer period with no WiFI / with account errors
- [ ] Test behaviour after app upgrade / installation
- [ ] Test behaviour after reboot
- [ ] Test behaviour after settings changes
- [ ] Test actual notifications
- [ ] Test discarding notifications
- [ ] Test snoozed notifications

johan12345:
> I wondered if there is a way to catch all exceptions in the service and present the ACRA notifications when the app is started for the next time.

Well, as all `Throwable`s are `Serializable`, we could save the exception(s) somewhere (in a file?), deserialize them when starting the app and then send them to ACRA.
Or we could send silent bug reports using ACRA directly from the service, which would have the disadvantage of not knowing the user's email address.
*BTW:* We can update ACRA to version 4.6 (seems to be [mainly bugfixes](https://github.com/ACRA/acra/wiki/ChangeLog)) and should remember to upgrade to 4.7 when the stable release is available because it [adds compatibility for Android 6.0](https://github.com/ACRA/acra#acra-v47)

For the testing todo: Snoozed notifications include both actually clicking the "later" button and dismissing or clicking the notification with the "Repeat notifications" setting activated

raphaelm:
Repeating after errors seems to work.
```
09-27 13:55:19.043 I/SyncAccountService(18804): SyncAccountService started
09-27 13:55:19.044 I/SyncAccountService(18804): SyncAccountService finished  with errors
09-27 14:58:14.686 I/SyncAccountService(18804): SyncAccountService started
09-27 14:58:14.687 I/SyncAccountService(18804): SyncAccountService finished  with errors
09-27 15:52:36.635 I/SyncAccountService(18804): SyncAccountService started
09-27 15:52:36.638 I/SyncAccountService(18804): SyncAccountService finished  with errors
09-27 17:07:30.439 I/SyncAccountService(18804): SyncAccountService started
09-27 17:07:30.439 I/SyncAccountService(18804): SyncAccountService finished  with errors
09-27 17:55:39.125 I/SyncAccountService(18804): SyncAccountService started
09-27 17:55:39.126 I/SyncAccountService(18804): SyncAccountService finished  with errors
```

raphaelm:
```
09-28 17:06:16.441 I/SyncAccountService(18804): SyncAccountService started
09-28 17:06:16.460 I/SyncAccountService(18804): Loading data for Account Account [id=3, library=Preetz, label=Unbenanntes Konto, name=123, cached=0]
09-28 17:06:27.202 I/SyncAccountService(18804): Loading data for Account Account [id=4, library=Brackenheim, label=Unbenanntes Konto, name=123, cached=0]
09-28 17:06:28.183 I/SyncAccountService(18804): Loading data for Account Account [id=6, library=Heidelberg_Uni, label=Unbenanntes Konto, name=123, cached=0]
09-28 17:06:32.404 I/SyncAccountService(18804): Loading data for Account Account [id=7, library=Wernau, label=Unbenanntes Konto, name=123, cached=0]
09-28 17:06:45.598 I/SyncAccountService(18804): SyncAccountService finished  successfully
09-29 17:07:31.700 I/SyncAccountService(18804): SyncAccountService started
09-29 17:07:31.732 I/SyncAccountService(18804): Loading data for Account Account [id=3, library=Preetz, label=Unbenanntes Konto, name=123, cached=1443496007603]
09-29 17:07:33.049 I/SyncAccountService(18804): Loading data for Account Account [id=4, library=Brackenheim, label=Unbenanntes Konto, name=123, cached=1443496009037]
09-29 17:07:37.240 I/SyncAccountService(18804): Loading data for Account Account [id=6, library=Heidelberg_Uni, label=Unbenanntes Konto, name=123, cached=1443496014025]
09-29 17:07:43.474 I/SyncAccountService(18804): Loading data for Account Account [id=7, library=Wernau, label=Unbenanntes Konto, name=123, cached=1443496032770]
09-29 17:07:46.703 I/ActivityManager(  868): Waited long enough for: ServiceRecord{7d36b4a u0 de.geeksfactory.opacclient.debug/de.geeksfactory.opacclient.reminder.SyncAccountService}
09-29 17:08:02.916 I/SyncAccountService(18804): SyncAccountService finished  successfully
```

Half-day seems to work (only one run per day is in the log because I always forget to keep the computer running logcat on overnight, we'll see whether I'll do it tonight).

raphaelm:
> Notifications displaying while the app is being updated or force-closed will disappear and not come again

Did you check out [ACTION_PACKAGE_REPLACED](https://developer.android.com/reference/android/content/Intent.html#ACTION_PACKAGE_REPLACED)? Starting from API 12, there is also [ACTION_MY_PACKAGE_REPLACED](https://developer.android.com/reference/android/content/Intent.html#ACTION_MY_PACKAGE_REPLACED).

If the user force-closes the app, I think we are out of luck.

johan12345:
> Did you check out ACTION_PACKAGE_REPLACED? Starting from API 12, there is also ACTION_MY_PACKAGE_REPLACED.

Oh, I'll try that.

> If the user force-closes the app, I think we are out of luck.

Yes. But only if they use the "Force Close" (or "Clear App Data") button in the App Info dialog, not when swiping away the app in the recents menu (AlarmManager alarms and Notifications shouldn't be removed in this case).

raphaelm:
> Yes. But only if they use the "Force Close" (or "Clear App Data") button in the App Info dialog, not when swiping away the app in the recents menu (AlarmManager alarms and Notifications shouldn't be removed in this case).

Well, that is absolutely okay.

johan12345:
I rebased this branch onto the changes from `master` (some conflicts in the APIs needed to be resolved) and am now going to continue testing.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/AccountFragment.java
```diff
@@ -847,76 +748,59 @@ public void onClick(View clicked) {
...
-                            holder.vStatusColor
-                                    .setBackgroundColor(getResources().getColor(R.color.date_warning));
-                        } else if (item.containsKey(AccountData.KEY_LENT_DOWNLOAD)) {
+                    if (item.getDeadline() != null) {
+                        if (item.getDeadline().equals(LocalDate.now()) ||
+                                item.getDeadline().isBefore(LocalDate.now())) {
+                            holder.vStatusColor.setBackgroundColor(
+                                    getResources().getColor(R.color.date_overdue));
+                        } else if (
+                                Days.daysBetween(item.getDeadline(), LocalDate.now()).getDays() <=
```
johan12345: This doesn't seem to be working. I don't know why, but it colors all items yellow, not only those that are in the warning period.

raphaelm: A breakpoint in the debugger should be able to tell you why ;)

johan12345: Yes, I tried that, but it didn't seem to be able to tell me what `Days.daysBetween` returned. I'll probably need to assign that to a variable.

raphaelm: Yes, or use the "Watch" feature of your debugger where you can watch the return value of single code snippets :)

johan12345:
Note to self: There seems to be a crash caused by a NullPointerException after adding a new account.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/AccountFragment.java
johan12345: Fixed in 99ea18e. The order of the arguments supplied to `Days.daysBetween` was incorrect.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/AccountFragment.java
johan12345: Fixed in feb5e53c

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/objects/AccountItem.java
johan12345: We could also provide both a String property and a `SearchResult.MediaType` property so we can choose what works for each system in the APIs and show the one that is available in the app (and optionally show an icon if the `SearchResult.MediaType` property is set).

raphaelm: I don't like the thought of maintaining two options, but the benefits seem to be worth it.

johan12345:
I tested rebooting the device. SyncAccountService started running again perfectly, but notifications I didn't dismiss before rebooting did not come back.

Maybe we should check somewhere if the notification is still displayed and otherwise set the `notified` property of the alarm to false again? This would also fix the same problem for notifications displayed while the app is being updated.

@raphaelm Any other idea?

opacclient/opacapp/src/main/res/values/strings.xml
raphaelm: How about „You might miss some notifications if you go without WiFi for a longer time“?

raphaelm:
Ok, let's try to get this done. As I see it, the ToDo at this point is:

- [ ] Think about re-showing undismissed notifications after a reboot (see your last comment)
- [ ] Handle ACTION_PACKAGE_REPLACED
- [ ] Test behaviour after app upgrade / installation

What options do we have here? We don't get notified if a user dismisses a notification, correct? Do we have a change of a well-informed decision whether to show a notification again?

- [ ] Make SyncAccountService only run if notifications are enabled

This should be trivial, but is important.

- [ ] Test behaviour after settings changes
- [ ] Test discarding notifications
- [ ] Test snoozed notifications

johan12345:
> We don't get notified if a user dismisses a notification, correct?

We do. An Intent gets called and it triggers `ReminderBroadcastReceiver`'s `notificationDeleted()` function.

opacclient/opacapp/src/main/res/values/strings.xml
johan12345: Okay. Added in c823240.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/objects/AccountItem.java
johan12345: I didn't do anything about this yet. Should I still add the `SearchResult.MediaType` property?

raphaelm: No, I think we'll first keep it as it is.

johan12345:
Sub-ToDos for "Test behaviour after settings changes":

- [x] Test behaviour after changing "repeat notifications" setting -> of course does not work for notifications that were already dismissed
- [ ] Test behaviour after changing "only WiFi" setting -> should apply for the next `SyncAccountService` run, does this need to be tested?
- [ ] Test behaviour after changing warning period
- [ ] Test behaviour after turning on/off notifications

johan12345:
I now completed the ToDos you listed (including rebase against the current master yesterday) and fixed/improved the behavior after settings changes in fdb1cd3. Anything else missing?

raphaelm:
I'll look deeper into it asap :)

PR #347, opened 2015-11-07
Potsdam_Uni and remove Potsdam

johan12345:
Potsdam_Uni now uses LBS and it is located on a different server.
Potsdam now blocks us based on User-Agent.


raphaelm:
Please change the pull request such that only Potsdam_Uni is affected and Potsdam is removed from the app completely. They have reasons for blocking as and asked to be removed, we have to respect this. See support ticket #2365608 for detailled information.

johan12345:
Oh, sorry, I forgot that Potsdam was the library that had requested to be removed. Now I changed the commit to remove Potsdam instead of fixing it.

PR #355, opened 2015-12-08
Re-enable ProGuard and update dependencies

johan12345:
We need to check if this ProGuard configuration fixes the problem on Wiko devices running Android 4.2.


raphaelm:
I just sent out a beta build to four affected users.

PR #392, opened 2016-03-29
WinBiap 4.4 and other fixes, new libraries

johan12345:
* Prolonging fixed for WinBiap 4.4 (tested in Memmingen)
* Getting cancelurl fixed for WinBiap 4.4 (tested using HTML sample, cancelling reservations itself not tested)
* New server for Baden and Wettingen
* Trier switched to OPEN
* Added libraries

For the first fix, I copied some code (essentially a RegEx) from the OPEN API implementation because both systems use Microsoft's ASP.NET framework and its "__doPostBack" Javascript mechanism. Should we instead transfer this code to a new function and place it in `BaseApi` or a new utility class?

### DIFF
--- a/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/WinBiap.java
+++ b/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/WinBiap.java
@@ -611,6 +611,19 @@ public class WinBiap extends BaseApi implements OpacApi {
             params.add(new BasicNameValuePair(kv.key(), kv.value()));
         }
 
+        if (lentPage.select("a[id$=ButtonBorrowChecked][href^=javascript]").size() > 0) {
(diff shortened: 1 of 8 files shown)


raphaelm:
As it is only 16 lines of code of duplicate so far, I don't really see a problem. As soon as it gets more, I'd love to see that in .utils. Would creating such a util class also simplify the current OPEN code significantly? If so, do it. If not, let's wait for the need.

johan12345:
Currently, this function could only be used at one position in both the OPEN and WinBiap APIs.

There is one more place in our internal Plus Edition-specific code for OPEN where we could use it, but don't have to as the arguments contained in the the `javascript:__doPostBack` link can also be (and are currently) obtained from CSS IDs.

raphaelm:
Then let's leave this at is for now.

PR #403, opened 2016-04-25
OPEN: improve search parsing

johan12345:
Support "comprehensiveItem"

### DIFF
--- a/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Open.java
+++ b/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Open.java
@@ -279,7 +279,7 @@ public class Open extends BaseApi implements OpacApi {
 
         int totalCount = Integer.parseInt(doc.select("span[id$=TotalItemsLabel]").first().text());
 
-        Elements elements = doc.select("div[id$=divMedium]");
+        Elements elements = doc.select("div[id$=divMedium], div[id$=divComprehensiveItem]");
         List<SearchResult> results = new ArrayList<>();
         int i = 0;
         for (Element element : elements) {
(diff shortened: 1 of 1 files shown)


opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Open.java
```diff
@@ -402,11 +406,22 @@ private String getCoverUrl(Element img) {
...
         // .BaseLibrary/StyleSheets/Images/Fallbacks/emptyURL.gif?4.2.0.0|a|
-        String url = null;
-        if (parts.length >= 2 && parts[0].equals("SetSimpleCover")) {
-            url = parts[2].replace("&amp;", "&");
+        for (int i = 0; i + 2 < parts.length; i++) {
+            if (parts[i].equals("SetSimpleCover")) {
+                String url = parts[i + 2].replace("&amp;", "&");
+                try {
+                    HttpURLConnection conn = (HttpURLConnection) new URL(url).openConnection();
+                    conn.setRequestMethod("HEAD");
```
raphaelm: While I'm gonna merge this for now, I really don't like this pattern of doing a lot of HEAD calls inside search(). We have something similar for another API, IIRC. This makes the search() call a lot slower and we *still* have to download the file if we want to display the cover. I think we need either put a concept of "alternative cover URLs" into the API so the frontend can do this asynchronously or download the file here completely and include the bitmap in the SearchResult.

johan12345: Yes, it would probably be better to have a possibility to provide multiple cover URLs. I would only save the Bitmap if it's absolutely required, because then we would need to care about things like translating between Android's Bitmap format and a platform-independent one and Java would determine which image compression formats we support etc.

PR #405, opened 2016-04-26
Fix off-by-one for Nuremberg fix.

ckeen:
The limit in string.split() introduced by commit
8dfbc947d19c960b4200ec88f555ee187962a540
triggers an array out of bounds exception. Also since we want the second
result, the limit is wrong anyway. The date given has the form
(DD.MM.YYYY - DD.MM.YYYY) with the start and end date of the lending
period. For the account listing the second date is the relevant one.


raphaelm:
Thanks a lot!

Issue #411, opened 2016-05-26
Add barcode scanning to the account edit view


raphaelm:
**Hey!** If you ever wanted to contribute to an Open Source project, but didn't know how, you just found a perfect place to start. We try to write this issue as instructive as possible, but as we're humans we almost certainly forgot to write something important in here. If you want to try to tackle this issue and get stuck at *any* point, please do not hesitate and ask us, either here on GitHub or via email at info@opacapp.de. We'll be very happy to help! 

We think this issue is very suitable to people who are just starting with Open Source or Android development in general and have never done it before. We'd like to ask people with wide experience in contributing to leave this issue to beginners. Thanks!

## How to get started

First of all, we recommend that you [install the app](https://play.google.com/store/apps/details?id=de.geeksfactory.opacclient) and play around with it so you know how it looks and feels. Next, you should use git to download this GitHub repository and open it in Android Studio. Press the green "Play" button to run it on your device. If this works, you have a working development setup, yay! If it doesn't, feel free to ask us for help!

## What this is about

When you first started the app, it asked you to create an 'account' for a library so the app knows which library's catalogue you want to search and how to log in there. If you do not actively use a public library, you probably created an anonymous account that can be used for search only. However, if you wanted to add login information, you'd need to type in a username and a password. In most libraries, this user name is an account number that is printed on your library membership card in form of a barcode.

We think it'd be cool if there was a "scan barcode" button to directly scan the barcode of the membership card and enter the number into the text field. This is the task of this issue.

For actually scanning the barcode, you'll need a barcode scanner app like [this one](https://play.google.com/store/apps/details?id=com.google.zxing.client.android). The app already contains code to communicate with these external apps.

## How to do this

I recommend forking this repository and creating a special git branch for this issue before you start. This will make it easier for you to submit a pull request later. If you've never done this before, there are a lot of good guides out there and you can always ask us questions.

* The first step would be to add the respective button. Luckily, we also have barcode buttons in the search form where you can take some inspiration. You find the account editing layout in [activity_account_edit.xml](https://github.com/opacapp/opacclient/blob/master/opacclient/opacapp/src/main/res/layout/activity_account_edit.xml) and a sample for a field with a barcode button in [searchfield_barcode.xml](https://github.com/opacapp/opacclient/blob/master/opacclient/opacapp/src/main/res/layout/searchfield_barcode.xml).
[...]

PR #413, opened 2016-05-30
Add a barcode scanner to new account page.

YKassim:
This PR addresses Issue #411


raphaelm:
Hello YKassim,

first of all, thank you very much for the pull request! I do have a few minor remarks before I merge this, but nothing major. Please tell me if something is unclear to you and I'll try to make it more clear.

First of all, you committed the ``.idea`` folder that Android Studio uses to store its settings. Our ``.gitignore`` file tries to prevent this, but you opened the whole repository in Android Studio instead of only the ``opacclient/`` subdirectory and our ignore pattern did not match. We did not document this properly, so this is *our fault*. Sorry for that. To fix it, please run the following command and then commit again. This will not delete the settings from your computer, but remove them from the repository.

    git rm --cached -r .idea/

Second, please format your code more coherently with the rest of the code (i.e. ``} else {`` instead of ``}else {``. Most of this will be done automatically for you by Android Studio if you select your changed code parts and press ``Ctrl``+``Alt``+``L`` (or ``Option``+``Command``+``L`` on Mac).

I'll add a few more comments directly inside the diff.

Cheers,
Raphael

opacclient/gradle.properties
```diff
@@ -11,9 +11,10 @@
...
-org.gradle.jvmargs=-Xmx2048m
+#org.gradle.jvmargs=-Xmx2048m
 
 # When configured, Gradle will run in incubating parallel mode.
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
-# org.gradle.parallel=true
\ No newline at end of file
+# org.gradle.parallel=true
+org.gradle.jvmargs=-XX\:MaxHeapSize\=256m -Xmx256m
```
raphaelm: Can you tell me why you removed the line above and changed this one? Did you run out of memory? We need to find a solution for this to work nicely on all developers computers.

We'll find some way to have this modified locally. You could try if it works when you revert these changes and insert your custom line into the ``local.properties`` file. I haven't tried that so far.

For now, it would be nice if you kept this change only to your local file. I suggest reverting this change and then using "git add --patch" to only partially add files to the commit to avoid committing it again.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/AccountEditActivity.java
```diff
@@ -50,20 +51,22 @@
...
 import de.geeksfactory.opacclient.apis.OpacApi.OpacErrorException;
+import de.geeksfactory.opacclient.barcode.BarcodeScanIntegrator;
 import de.geeksfactory.opacclient.objects.Account;
 import de.geeksfactory.opacclient.objects.Library;
 import de.geeksfactory.opacclient.reminder.ReminderHelper;
 import de.geeksfactory.opacclient.storage.AccountDataSource;
 import de.geeksfactory.opacclient.utils.ErrorReporter;
 
-public class AccountEditActivity extends AppCompatActivity {
+public class AccountEditActivity extends AppCompatActivity implements SearchFragment.Callback {
```
raphaelm: You don't need to implement this interface, just strip it.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/AccountEditActivity.java
```diff
@@ -189,6 +201,36 @@ private void delete() {
...
+                account.setName(scanResult.getContents());
+                etName.setText(account.getName());
+            }
+
+
+        }
+    }
+
+    @Override
+    public void scanBarcode() {
```
raphaelm: You don't really need this method as it won't be called from outside, would you mind just moving this code up into your onClick handler?

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/AccountEditActivity.java
```diff
@@ -189,6 +201,36 @@ private void delete() {
     }
 
     @Override
+    public void onActivityResult(int requestCode, int resultCode, Intent idata) {
+        super.onActivityResult(requestCode, resultCode, idata);
+
+        //TODO: Rewrite this for the new SearchField implementation
```
raphaelm: Could you remove this comment? It is not relevant here :) Thanks!

opacclient/opacapp/src/main/res/layout/searchfield_barcode.xml
```diff
@@ -33,7 +33,8 @@
             android:contentDescription="@string/barcodes"
             app:srcCompat="@drawable/ic_barcode"
             tools:ignore="MissingPrefix"
-            tools:src="@drawable/ic_barcode"/>
+            tools:src="@drawable/ic_barcode"
+            android:clickable="false" />
```
raphaelm: Why did you change this attribute to false? I don't think that makes any sense.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/AccountEditActivity.java
```diff
@@ -189,6 +201,36 @@ private void delete() {
...
+                .parseActivityResult(requestCode, resultCode, idata);
+        if (resultCode != RESULT_CANCELED && scanResult != null) {
+            if (scanResult.getContents() == null) {
+                return;
+            }
+            else if (scanResult.getContents().length() < 3) {
+                return;
+            }else {
+
+                account.setName(scanResult.getContents());
```
raphaelm: You can remove this line, this will be done automatically if the user saves the changes.

raphaelm:
I added all of my comments for now. Please do not feel overwhelmed by them, this is perfectly normal and is not meant to critizise you, just to keep this codebase clean for the future :)

opacclient/gradle.properties
YKassim: It refuses to build and gives this error:

> Gradle sync failed: Unable to start the daemon process.
           This problem might be caused by incorrect configuration of the daemon.
           For example, an unrecognized jvm option is used.
           Please refer to the user guide chapter on the daemon at https://docs.gradle.org/2.10/userguide/gradle_daemon.html
           Please read the following process output to find out more:
           -----------------------
           Error occurred during initialization of VM
           Could not reserve enough space for 2097152KB object heap
It's solved when i add: org.gradle.jvmargs=-XX\:MaxHeapSize\=256m -Xmx256m to gradle.properties. 

The error persists when i add it to local.properties.

johan12345: Could you try to put your `org.gradle.jvmargs=-XX\:MaxHeapSize\=256m -Xmx256m` setting in a `gradle.properties` file in your home directory ("/home/username" on Linux, "C:\Users\username" on Windows), leaving the `gradle.properties` in the project directory as it was before? The file from your home directory should be able to override the settings.

raphaelm:
Your ``.idea/`` is gone from the commit now but somehow, your Android Studio still accidentally deleted our code style config etc. from the repository. This happend to another contributor as well and we believe this is our fault, not yours. I opened issue #420 to fix this.

Meanwhile, I manually merged your changes into the repository (I squashed them into one single commit and removed the said problem). They are now in the master branch as of 20232db and will be included in our next app release. Thank you very much, your contribution is highly appreciated!

YKassim:
Thank You, This was a very good learning experience.

Issue #416, opened 2016-06-01
Add Opac4 API and few Polish libraries


thesebas:
There is number of libraries with opac4 api:
* http://185.43.138.133:8080/Opac4/faces/Szukaj.jsp
* http://wbpg.org.pl:8080/Opac4/faces/Szukaj.jsp
* http://78.31.92.231/Opac4/faces/Szukaj.jsp
* http://katalog.mbp.slupsk.pl/Opac4/faces/Szukaj.jsp
* http://sowa.aps.edu.pl:8080/Opac4/faces/Szukaj.jsp
* http://mbp.tczew.pl:8080/Opac4/faces/Szukaj.jsp
* http://www.biblioteka.lazarski.pl/Opac4/faces/Szukaj.jsp
* http://biblioteka.wejherowo.pl:81/Opac4/faces/Szukaj.jsp
* http://libra.imir.agh.edu.pl:8080/Opac4/faces/Szukaj.jsp
* http://83.3.164.226:8080/Opac4/faces/Szukaj.jsp
* http://80.51.108.66:8080/Opac4/faces/Szukaj.jsp
* http://95.50.163.82/Opac4/faces/Szukaj.jsp

PR #417, opened 2016-06-02
Opac4 API implementation + example libraries

thesebas:
This pull request is just placeholder for #416, will try to implement and let you know when it's ready ASAP :)


raphaelm:
Uh, awesome! Excited to see this coming :)

johan12345:
Very nice that you are starting to implement this :)

Just a small remark: It seems the library software that hosts these OPACs is called [PATRON](http://www.patron.pl/?lang=en), so I think it would be better to change both the name of the API class and the keyword for the JSON files to `Patron` and `patron` , respectively (or maybe `Patron4` and `patron4`). @raphaelm, do you agree?

thesebas:
`Patron`/`patron` sounds good.

thesebas:
What about new `SearchResult.MediaType`s? I'd like to add two quite common I think: "Music Sheet", "Old Print" and there are other two exotic I'm not interested in (something called "Regional Bibliography" and "Social Life Documentation"). Is it just matter of adding them as enums or more dependent code?

johan12345:
"Music Sheet" should be `SCORE_MUSIC`. "Old print" is something that could be added. I don't think we need the two exotic ones either, you can just assign `UNKNOWN` to those.

When you add something like `OLD_PRINT` to the `SearchResult.MediaType` enum, you also need to modify the [`getResourceByMediaType`](https://github.com/opacapp/opacclient/blob/master/opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/ResultsAdapter.java#L50) function so that the app knows which icon to display for this media type. Do you have a suggestion for an icon to use? Maybe something like the letter "A" in an old font? The other icons we use are mostly from [materialdesignicons.com](https://materialdesignicons.com/), but I have not found something appropriate there yet.

raphaelm:
Hi @thesebas, just for your information, we just merged a major change into the ``master`` branch: We now no longer hold the JSON files inside the git repository but on a webserver. The goal is that we can update them easily without issuing an app update every few weeks. For now, you can go an as you currently do, but if you are done you'll need to remove the JSON files from the PR and we need to add them to the webservice.

thesebas:
Hi @raphaelm, I've rebased my pull request on current master, I'd like to remove json files now but to do tests if it works I need to have those json files in your webservice, probably I should create PR in https://github.com/opacapp/opacapp-config-files, is it enough?

johan12345:
You should be able to test it without us adding the JSON files to the webservice. Maybe you need to copy [this file](https://github.com/opacapp/opacapp-config-files/blob/master/last_library_config_update.txt) to your `src/main/assets` folder.

Nevertheless, I have already added your three JSON files to the webservice, so they should also appear when you run `gradlew downloadJson`.

For additional libraries, you can create a PR at opacapp-config-files, we will add them then.

raphaelm:
FYI, I disabled them in the webservice again to avoid them being included in our 5.0.3 release :)

PR #444, opened 2016-08-13
[WIP] Automatically update library configuration using opacdata API

johan12345:
We have built a web application running under `info.opacapp.net` that (amongst other things) stores all the library configuration files. We want to use its API to make it possible to update the configuration files without requiring an app update. This Pull Request implements two components:

- `JsonFilesTask`: A Gradle task that downloads the current version of the configuration files to include in an App build and saves the date where the files were last updated to a text file. This can be run using `gradlew downloadJson`.
- `LibraryConfigUpdateService`: An Android service that asks the API if configuration files were updated and downloads them to the local storage

**TO DO:**
*app*
- [x] run service regularly
- [x] actually use the updated JSON files in the app
- [x] handle `"_active": false`
- [x] replace Toast with ProgressDialog when running service from settings
- [x] clear downloaded JSON files after app update
- [x] write some Unit Tests
- [x] write documentation about this
- [ ] delete JSON files from this GitHub repository and add them to `.gitignore`

*opacdata*
- [ ] update configuration stored in opacdata with newest changes from this GitHub repository
- [x] implement API to only show files compatible with the current app version
- [ ] automatically change `lastUpdate` when configuration is edited (not that important)
- [ ] optionally create another repository for the JSON files and hook it up to opacdata preserve the possibility to contribute?


johan12345:
> handle `"_active": false`
We currently only send libraries with `"active": true` from the API. This makes it impossible for the app to know when a library was removed.

johan12345:
> run service regularly

How should we handle this? Should we run the service in the background every day or so? Or just when starting the app?

opacclient/opacapp/src/main/res/values-de/strings.xml
```diff
@@ -288,4 +289,15 @@
     <string name="starred_removed_undo">Rückgängig</string>
     <string name="print">Drucken</string>
     <string name="print_wait">Bitte warten, bis die Details geladen wurden!</string>
+    <string name="preference_category_library_config">Bibliotheksdaten</string>
+    <string name="preference_update_library_config">Konfiguration für Bibliotheken aktualisieren</string>
+    <string name="preference_reset_library_config">Konfiguration zurücksetzen</string>
```
raphaelm: I believe "Konfiguration" is misleading, as it sounds like the user could change it. Maybe just "Bibliotheksdaten aktualisieren/zurücksetzen"?

opacclient/opacapp/src/main/res/values-de/strings.xml
```diff
@@ -288,4 +289,15 @@
...
+    <string name="preference_reset_library_config_desc">zum Stand des letzten App-Updates</string>
+    <string name="library_config_no_updates">Keine aktualisierte Konfiguration für Bibliotheken gefunden.</string>
+    <string name="updating_library_config">Aktualisiere Konfiguration für Bibliotheken</string>
```
raphaelm: Similarly "Aktualisiere Bibliotheksdaten"

chandi1988:
Uu

chandi1988 (approved):

Issue #446, opened 2016-08-14
Replace EditTexts in AccountEditActivity with TextInputLayouts


johan12345:
**Hey!** If you ever wanted to contribute to an Open Source project, but didn't know how, you just found a perfect place to start. We try to write this issue as instructive as possible, but as we're humans we almost certainly forgot to write something important in here. If you want to try to tackle this issue and get stuck at *any* point, please do not hesitate and ask us, either here on GitHub or via email at info@opacapp.de. We'll be very happy to help!

We think this issue is very suitable to people who are just starting with Open Source or Android development in general and have never done it before. We'd like to ask people with wide experience in contributing to leave this issue to beginners. Thanks!

## How to get started

First of all, we recommend that you [install the app](https://play.google.com/store/apps/details?id=de.geeksfactory.opacclient) and play around with it so you know how it looks and feels. Next, you should use git to download this GitHub repository and open the folder `opacclient/` from within the repository (not the repository itself) in Android Studio.

Because of a [Bug in Android Studio](https://code.google.com/p/android/issues/detail?id=172347), some configuration files in the `.idea` folder might be deleted after opening the project for the first time. Run ``git checkout -- opacclient/.idea/`` in the repository to revert this.

Press the green "Play" button to run it on your device. If this works, you have a working development setup, yay! If it doesn't, feel free to ask us for help!

## What this is about

When you start the app and set up a library account, you can input details of your account, namely a label, your username or library membership ID and password.

We would like to replace the `EditText`s for these three inputs with the [`TextInputLayout`](https://developer.android.com/reference/android/support/design/widget/TextInputLayout.html) from the Android Design Support Library.

## How to do this

We recommend forking this repository and creating a special git branch for this issue before you start. This will make it easier for you to submit a pull request later. If you've never done this before, there are a lot of good guides out there and you can always ask us questions.

The XML layout file you need to look at is [`activity_account_edit.xml`](https://github.com/opacapp/opacclient/blob/master/opacclient/opacapp/src/main/res/layout/activity_account_edit.xml). Replacing the three `EditTexts` with `TextInputLayouts` should be quite straightforward. Make sure that the `ImageView` `ivBarcode` stays on the right side of the input for the username/membership ID.

If you should need to change something (such as View IDs and class imports) in the Java code of the Activity, it is located [here](https://github.com/opacapp/opacclient/blob/master/opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/AccountEditActivity.java).

[...]

Issue #445, opened 2016-08-14
Improve printing feature with CSS


johan12345:
**Hey!** If you ever wanted to contribute to an Open Source project, but didn't know how, you just found a perfect place to start. We try to write this issue as instructive as possible, but as we're humans we almost certainly forgot to write something important in here. If you want to try to tackle this issue and get stuck at *any* point, please do not hesitate and ask us, either here on GitHub or via email at info@opacapp.de. We'll be very happy to help!

We think this issue is very suitable to people who are just starting with Open Source or Android development in general and have never done it before. We'd like to ask people with wide experience in contributing to leave this issue to beginners. Thanks!

**Solving this issue requires very little experience in Java/Android programming, but you should be familiar with HTML and CSS.**

## How to get started

First of all, we recommend that you [install the app](https://play.google.com/store/apps/details?id=de.geeksfactory.opacclient) and play around with it so you know how it looks and feels. Next, you should use git to download this GitHub repository and open the folder `opacclient/` from within the repository (not the repository itself) in Android Studio.

Because of a [Bug in Android Studio](https://code.google.com/p/android/issues/detail?id=172347), some configuration files in the `.idea` folder might be deleted after opening the project for the first time. Run ``git checkout -- opacclient/.idea/`` in the repository to revert this.

Press the green "Play" button to run it on your device. If this works, you have a working development setup, yay! If it doesn't, feel free to ask us for help!

## What this is about

When you start the app and set up a library account, you can search this library's catalog, view the search result list and click on the results to get detailed information about a single item, including things like where it is located in the library. Pull Request #440 (see also issue #431) added a printing feature to the app, which means that these details can be sent to a printer using the print button (or overflow menu item on narrow screens) on devices with Android >= 4.4. The print feature is based on an HTML page that is generated by the app.

The printing feature is not yet included in the current version of the app on the Play Store, so you have to install the app from Android Studio if you want to try it.

This issue is about making the printed output a little nicer by applying some CSS to the generated HTML. We have listed some suggestions below, but if you have more ideas, feel free to ask us here and implement them afterwards.

## How to do this

We recommend forking this repository and creating a special git branch for this issue before you start. This will make it easier for you to submit a pull request later. If you've never done this before, there are a lot of good guides out there and you can always ask us questions.

[...]

PR #449, opened 2016-09-07
Replaced EditTexts in AccountEditActivity with TextInputLayouts (#446)

NPuhlmann:
As in Issue #446, I replaced the EditTexts.

### DIFF
--- a/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Bibliotheca.java
+++ b/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Bibliotheca.java
@@ -1103,7 +1103,9 @@ public class Bibliotheca extends BaseApi {
             String header = headerCell.text();
             headersList.put(header);
             if (headers_reservations.has(header)) {
-                copymap.put(headers_reservations.getString(header), i);
+                if (!headers_reservations.isNull(header)) {
+                    copymap.put(headers_reservations.getString(header), i);
+                }
             } else {
                 unknownHeaders.put(header);
             }
--- a/opacclient/opacapp/src/main/assets/bibs/Balve.json
+++ b/opacclient/opacapp/src/main/assets/bibs/Balve.json
@@ -1,0 +1,16 @@
+{
+  "account_supported": true,
+  "api": "winbiap",
+  "city": "Balve",
+  "country": "Deutschland",
+  "data": {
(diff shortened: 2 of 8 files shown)


raphaelm:
Hi @NPuhlmann, thanks a lot! I'm not sure I'll be able to try this out this week, but I will definitively look into it next week.

johan12345:
Hi @NPuhlmann,

it seems like you have just replaced the `EditText`s with `android.support.design.widget.TextInputEditText`s, which has no visual effect until you also surround them with a `android.support.design.widget.TextInputLayout`, as written in the [documentation](https://developer.android.com/reference/android/support/design/widget/TextInputLayout.html).

Could you fix this? After adding the TextInputLayouts, you can probably also remove the label TextViews (they have the same text as the hint in the TextInputLayouts).

NPuhlmann:
Hi @johan12345 and @raphaelm, 
I replaced my mistake from above and now you can see the visual effects :) I also did #445 and hope you like it. I was not able to set it up with your solution but with floating the table elements. Added some break in page and break not in page rules for a better structure. I have not found a view with some tables and ul´s so maybe the float has to be cleared at the start of the ul´s?

raphaelm:
Hi @NPuhlmann, there is some confusion going on here. With doing the CSS changes in your ``master`` branch after you already created this PR on the ``master`` branch, this PR now contains the TextInputLayout *and* the CSS changes. With you now pulling another branch, this PR asks us to include changes that are already included. Don't worry, git is a mystery to everyone in the beginning. I'll see that I review this today and tomorroy and I'll fix this up for you by merging the appropriate commits. Until then, it'd be the best idea to leave this branch as it is.

For the future, the best workflow is to create a so-called feature branch on top of master before you make any changes and then do the changes for every PR seperately in the respective branch. 

If you have any questions, just drop me aline!

NPuhlmann:
Hi @raphaelm, thanks for this information. I thought the build error causes from a wrong build file so i checked my branches for this and pushed everything back to my fork. Ok, next time I will set up a branch for each feature. Thanks for the hint! :)

raphaelm:
Hi @NPuhlmann, I manually integrated your changes into our ``master`` branch. I like the idea to use ``page-break-inside: avoid``, but I decided against the ``page-break-before: always`` as it causes more paper to be used on items with very few details.

Thank you very much for the patches!

PR #450, opened 2016-09-11
Rebuild account lists using RecyclerViews, add media type and cover

johan12345:


johan12345:
I think this is ready for testing and review.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/storage/AccountDataSource.java
```diff
@@ -284,6 +288,8 @@ private void setAccountItemAttributes(AccountItem item, ContentValues cv, long a
         putOrNull(cv, "format", item.getFormat());
         putOrNull(cv, "itemid", item.getId());
         putOrNull(cv, "status", item.getStatus());
+        putOrNull(cv, "cover", item.getCover());
+        putOrNull(cv, "mediatype", item.getMediaType().toString());
```
raphaelm: This crashed for me repeatedly on APIs where getMediaType returns null, e.g. the Test API or Adis.

raphaelm:
The list still doesn't look very nice if there are no media types set. How do we fix this?

johan12345:
Done :)

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/storage/AccountDatabase.java
```diff
@@ -68,12 +68,13 @@ public void onCreate(SQLiteDatabase db) {
                         "account integer," + "title text," + "author text," + "format text," +
                         "itemid text," + "status text," + "barcode text," + "deadline text," +
                         "homebranch text," + "lending_branch text," + "prolong_data text," +
-                        "renewable integer," + "download_data text," + "ebook integer" + ");");
+                        "renewable integer," + "download_data text," + "ebook integer" +
+                        "mediatype text," + "cover text," + ");");
```
raphaelm: The ``,`` at the end breaks the install.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/storage/AccountDatabase.java
```diff
@@ -68,12 +68,13 @@ public void onCreate(SQLiteDatabase db) {
...
         db.execSQL("create table " + "accountdata_reservations (" +
                 "id integer primary key autoincrement," + "account integer," + "title text," +
                 "author text," + "format text," + "itemid text," + "status text," + "ready text," +
-                "expiration text," + "branch text," + "cancel_data text," + "booking_data text" +
-                ");");
+                "expiration text," + "branch text," + "cancel_data text," + "booking_data text," +
+                "mediatype text," + "cover text," + ");");
```
raphaelm: Same here :)

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/storage/AccountDatabase.java
johan12345: fixed

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/storage/AccountDatabase.java
johan12345: fixed

raphaelm:
The animation is a lot smoother now :) I just added two other commits, the one containing a crash that I occured multiple times. Still, there are some issues I wasn't able to solve immediately:

First, when scrolling multiple times, the covers are not always hidden properly:
device-2016-10-09-151846

Second, the lines between the items are sometimes missing during scrolling and are missing for the reservations all the time:
device-2016-10-09-153049

Third, when opening a long detail dialog and scrolling down and then opening a short one, it is scrolled as well:
device-2016-10-09-153058

opacclient/opacapp/src/main/res/layout/content_accountitem_detail.xml
```diff
@@ -90,7 +90,7 @@
                     android:layout_width="match_parent"
                     android:layout_height="wrap_content"
                     android:padding="16dp"
-                    android:visibility="@{item.author != null ? View.VISIBLE : View.GONE}">
+                    android:visibility="@{(item.author != null &amp;&amp; !item.author.equals(&quot;&quot;)) ? View.VISIBLE : View.GONE}">
```
johan12345: Shouldn't we ensure that `author == null || !author.equals("")` in the APIs (e.g. in the Unit Tests, similarly for all the other properties)? Always checking for both `null` and empty Strings makes code like this somewhat ugly :/

raphaelm: You're right (although that should be a rare case limited to the weird string parsing in Biber which I used to test). Should've fixed it in Biber, will fix it in both places.

PR #454, opened 2016-09-30
New intro

johan12345:
Creating a new, Material Design intro screen (with three pages) for the app


raphaelm:
I totally love it! :) Two questions:

* Do you see an easy way to replace the "skip" icon by a text "SKIP" or an X symbol?
* On the third page, the only call to action in the text is to suggest a library and the continue button is a tick. How do you think about hiding the tick and placing a "SELECT A LIBRARY" button below the text as the primary action of the slide?

johan12345:
> Do you see an easy way to replace the "skip" icon by a text "SKIP" or an X symbol?

I did that on the `custom_layout branch` of [my `material-intro` repo](https://github.com/johan12345/material-intro/commits/custom_layout). This also relates to https://github.com/HeinrichReimer/material-intro/issues/60.

> On the third page, the only call to action in the text is to suggest a library and the continue button is a tick. How do you think about hiding the tick and placing a "SELECT A LIBRARY" button below the text as the primary action of the slide?

Done.

raphaelm:
Does not build for me: ``Error:(7, 44) No resource found that matches the given name (at 'dialog_width' with value '@dimen/dialog_width_major').``

raphaelm:
TODO, as I see it:

* Can we do something easy about the fact that the third page offers scrolling?
* How do we deal with phones in landscape mode?

johan12345:
> Can we do something easy about the fact that the third page offers scrolling?

All pages offer scrolling if there's not enough space. We could turn scrolling off entirely and then make the images scale appropriately (down to 0 on very small screens or with very long texts).

> How do we deal with phones in landscape mode?

I have added a landscape layout (for both phones and tablets).

PR #455, opened 2016-10-03
Improve AccountEditActivity UX

johan12345:
The Text inputs for username and password were labelled as "optional", which caused confusion for some users. Therefore, I added two RadioButtons to select whether the account is anonymous or has credentials. When the former is selected, the two inputs are hidden.


raphaelm:
I believe that, if we have account support for the library, that should be the default.

johan12345:
fixed :)

Issue #457, opened 2016-10-20
Crash in F-Droid builds >=5.0


github-user-1:
I just upgraded to version 5.0.2 via F-Droid earlier today (Fairphone 2). Since then the app doesn't start anymore and I constantly get the Android-notification that the app crashed...

raphaelm:
Does the notification offer you to send in a crash report? If so, please do that!

github-user-1:
Unfortunately, not - this is the notification I get:

![screenshot_2016-10-20-18-56-07](https://cloud.githubusercontent.com/assets/6796538/19569917/231825b6-96e7-11e6-9651-1b4d8107c6d0.png)

If I click on this notification, I get:

![screenshot_2016-10-20-18-56-47](https://cloud.githubusercontent.com/assets/6796538/19569944/49d903d2-96e7-11e6-90d8-0bfd5f1d18d9.png)

Same thing happens if I just try to close the notification (swiping left of right). Android seems to try to start the app but always fails to do so.

uvok:
Even after uninstall and reinstall, thus deleting app data, app crashes on startup. 
I could get a logcat. It's either one of these. 

``` 10-20 19:55:16.669 E/ACRA    (20166): ACRA caught a RuntimeException for de.geeksfactory.opacclient
10-20 19:55:16.669 E/ACRA    (20166): java.lang.RuntimeException: Unable to create application de.geeksfactory.opacclient.OpacClient: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String org.joda.time.DateTime.toString()' on a null object reference
10-20 19:55:16.669 E/ACRA    (20166): 	at de.geeksfactory.opacclient.OpacClient.onCreate(OpacClient.java:386)```

```10-20 19:55:16.449 E/ACRA    (20143): de.geeksfactory.opacclient fatal error : Unable to create application de.geeksfactory.opacclient.OpacClient: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String org.joda.time.DateTime.toString()' on a null object reference
10-20 19:55:16.449 E/ACRA    (20143): java.lang.RuntimeException: Unable to create application de.geeksfactory.opacclient.OpacClient: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String org.joda.time.DateTime.toString()' on a null object reference
[...]

PR #458, opened 2016-10-21
Don't crash when last_library_config_update.txt is missing from the APK

johan12345:
In release builds this situation normally should not appear (but it currently does with F-Droid builds, see #457). But this is also helpful for making debug builds possible without running `downloadJson` first.


raphaelm:
I don't believe this is a sufficient fix. Imagine the following:

* User installs 5.0.3 from F-Droid without any data in it
* App downloads data via the service, user creates a library account
* User installs 5.0.4 from F-Droid. Using the account will fail (in what way?) because the app currently does not use downloaded data that has been downloaded with a different version.

johan12345:
Improved with #459.

PR #464, opened 2016-12-12
Add id retrieval for Stadtbibliothek München in Adis

sw-dev1:
@raphaelm: As requested earlier today, this change fills the item id field in AccountData with the 14 digit library id, parsed from the title in the html. Works for the Stadtbibliothek München.


raphaelm:
Thanks!

Issue #473, opened 2017-01-11
Account editing view: Add visibility toggle to password input


johan12345:
**Hey!** If you ever wanted to contribute to an Open Source project, but didn't know how, you just found a perfect place to start. We try to write this issue as instructive as possible, but as we're humans we almost certainly forgot to write something important in here. If you want to try to tackle this issue and get stuck at _any_ point, please do not hesitate and ask us, either here on GitHub or via email at info@opacapp.de. We'll be very happy to help!

We think this issue is very suitable to people who are just starting with Open Source or Android development in general and have never done it before. We'd like to ask people with wide experience in contributing to leave this issue to beginners. Thanks!

## How to get started

First of all, we recommend that you [install the app](https://play.google.com/store/apps/details?id=de.geeksfactory.opacclient) and play around with it so you know how it looks and feels. Next, you should use git to download this GitHub repository and open the folder `opacclient/` from within the repository (not the repository itself) in Android Studio.

Because of a [Bug in Android Studio](https://code.google.com/p/android/issues/detail?id=172347), some configuration files in the `.idea` folder might be deleted after opening the project for the first time. Run `git checkout -- opacclient/.idea/` in the repository to revert this.

Before running the app, you should run the `downloadJson` Gradle task which loads the configuration files for all the libraries in the app from our server. You can either do this using Android Studio's *Gradle Tool Window* or by running `gradlew downloadJson` on the command line.

Press the green "Play" button to run it on your device. If this works, you have a working development setup, yay! If it doesn't, feel free to ask us for help!

## What this is about

When you start the app and set up a library account, you can input details of your account, namely a label, your username or library membership ID and password. This issue is about the password input field.

The Android Support Library supports adding a button to password input fields that allows you to toggle the password between the normal "disguised" state and a visible state. An example of this can be seen [here](https://material.io/guidelines/components/text-fields.html#text-fields-password-input).

We would like you to add this button to our password input. As a security measure, it should only be shown when an account is first created or when there is no password currently set for the account. This makes sure that the password can not be revealed later by a malicious user opening the account editing view again.

## How to do this

We recommend forking this repository and creating a special git branch for this issue before you start. This will make it easier for you to submit a pull request later. If you've never done this before, there are a lot of good guides out there and you can always ask us questions.

[...]

PR #476, opened 2017-03-06
Added PasswordToggle in case that Password wasn´t set before. #473

NPuhlmann:
As in Issue #473 there now is an toggle button for the password if it was not set before.


raphaelm:
Thank you very much! :) Could you just fix up the formatting of the change? Android Studio should do that automatically for you if you select your change and press Ctrl+Alt+L.

NPuhlmann:
I´m working with Linux Mint so CTRL ALT L locks the screen :P But Code > Reformat Code worked also ;)

raphaelm:
I think it was Ctrl Shift L or something like this, my bad :D thanks anyway!

PR #478, opened 2017-03-21
Add buttons to get the app

Poussinou:
Buttons with links to fdroid and google play store


raphaelm:
Hi, thanks for the idea :)

Issue #484, opened 2017-08-02
Export and import favorites list


johan12345:
**Hey!** If you ever wanted to contribute to an Open Source project, but didn't know how, you just found a perfect place to start. We try to write this issue as instructive as possible, but as we're humans we almost certainly forgot to write something important in here. If you want to try to tackle this issue and get stuck at _any_ point, please do not hesitate and ask us, either here on GitHub or via email at info@opacapp.de. We'll be very happy to help!

We think this issue is very suitable to people who are just starting with Open Source or Android development in general and have never done it before. We'd like to ask people with wide experience in contributing to leave this issue to beginners. Thanks!

## How to get started

First of all, we recommend that you [install the app](https://play.google.com/store/apps/details?id=de.geeksfactory.opacclient) and play around with it so you know how it looks and feels. Next, you should use git to download this GitHub repository and open the folder `opacclient/` from within the repository (not the repository itself) in Android Studio.

Before running the app, you should run the `downloadJson` Gradle task which loads the configuration files for all the libraries in the app from our server. You can either do this using Android Studio's *Gradle Tool Window* or by running `./gradlew downloadJson` on the command line.

Press the green "Play" button to run it on your device. If this works, you have a working development setup, yay! If it doesn't, feel free to ask us for help!

## What this is about

When you start the app and set up a library account (just choose any library and select "anonymous account", you do not need a password), you can search for items in that library. If you click one of the results to go to the detail view, there is a "star" icon to add this item to your favorites list.

The favorites list is only saved locally in the app - so when users switch to a new phone, the items that were saved to their favorites list will be gone (if you don't use any backup solution). This is why we would like you to add a simple export and import functionality to the app.

We already have a feature that allows to share the items on the favorites list as a human-readable text (including the book title and the link to the library catalog), but we think that for an export/import functionality, the better data format would be a JSON file.

## How to do this

We recommend forking this repository and creating a special git branch for this issue before you start. This will make it easier for you to submit a pull request later. If you've never done this before, there are a lot of good guides out there and you can always ask us questions.

The favorites list is displayed by the `StarredFragment` class, which also contains the [implementation of our current sharing feature](https://github.com/opacapp/opacclient/blob/master/opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/StarredFragment.java#L228). We would like you to create a similar function that exports the data to a JSON file and allows the user to choose a location to save it to. You can add a menu item to the Action Bar that will trigger this export feature.

[...]

PR #485, opened 2017-08-21
Implement export-import favorites list

kedarmp:
Import export implemented as per issue #484


opacclient/opacapp/src/main/AndroidManifest.xml
```diff
@@ -7,6 +7,8 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
+    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
```
raphaelm: Are you sure we need this? It wouldn't be a killer, but since @johan12345 specifically wrote
> We think the best way for this is using Android's Storage access framework, because it does not require additional permissions and already provides you with a file chooser dialog

I want to double check.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/StarredFragment.java
```diff
@@ -157,15 +184,21 @@ public void onItemClick(AdapterView<?> parent, View view,
...
     @Override
     public boolean onOptionsItemSelected(android.view.MenuItem item) {
         if (item.getItemId() == R.id.action_export) {
-            export();
+            share();
+            return true;
+        } else if (item.getItemId() == R.id.action_export_to_storage) {
+            exportToStorage();
+            return true;
+        } else if (item.getItemId() == R.id.action_import_to_storage) {
```
raphaelm: import_from would probably be a more fitting name

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/StarredFragment.java
```diff
@@ -256,6 +290,166 @@ protected void export() {
...
+        //android 4.4+; use Storage Access Framework
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.KITKAT) {
+            intent = new Intent(Intent.ACTION_CREATE_DOCUMENT);
+            intent.addCategory(Intent.CATEGORY_OPENABLE);
+            // Create a file with the requested MIME type.
+            intent.setType("application/json");
+            intent.putExtra(Intent.EXTRA_TITLE, "share.json");
+            startActivityForResult(intent, REQUEST_CODE_EXPORT);
+        } else {        // <android 4.4; share raw json file
+            if(Environment.getExternalStorageState()==Environment.MEDIA_MOUNTED) {
```
raphaelm: Could you run "Format code" in Android Studio from time to time to have the code formatted coherently? :)

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/StarredFragment.java
```diff
@@ -256,6 +290,166 @@ protected void export() {
...
+                File dest = new File(getActivity().getExternalCacheDir(),"starred_export.json");
```
raphaelm: Do we really need the cached file? Couldn't we just pass the JSON content as ``Intent.EXTRA_TEXT`` instead? Intents can handle data on the order of a couple of hundreds of kilobytes which should be enough, at least for an old-device fallback solution that we don't really want to support. @johan12345 what do you think?

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/StarredFragment.java
```diff
@@ -256,6 +290,166 @@ protected void export() {
...
+                showExportError();
+            }
+        } else if (requestCode == REQUEST_CODE_IMPORT && resultCode == Activity.RESULT_OK) {
+            Uri uri = data.getData();
+            try {
+                StarDataSource dataSource = new StarDataSource(getActivity());
+                InputStream is = getActivity().getContentResolver().openInputStream(uri);
+                if(is!=null) {
+                    BufferedReader reader = new BufferedReader(new InputStreamReader(is, "UTF-8"));
+                    String line = reader.readLine();
```
raphaelm: We should also be able to parse JSON that is distributed over more than one line, so we should first consume the reader into a string, not just the first line.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/StarredFragment.java
kedarmp: I agree!

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/StarredFragment.java
kedarmp: We could, but then the onus of actually creating/storing the JSON file would now be on the owner since with `Intent.EXTRA_TEXT`, we would have to pass the raw JSON string, which seems pointless from a user's POV. For example, it would be easy for a user to accidentally edit this content before he saves it somewhere. Which is why I believe we need the cached file.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/StarredFragment.java
kedarmp: My assumption was that the app wouldn't come across a JSON file that does not conform to its (private) format (since the app itself creates the JSON in the first place).

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/StarredFragment.java
raphaelm: The problem is, that we then need a new permission, only because of users on Android <4.4 and its especially those users where new permissions break auto-updates so users won't update the app :\

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/StarredFragment.java
raphaelm: Well, nothing is preventing a user from modifying the file while transfering it to a new device and since JSON is a kind-of-standard, it would be great not to impose arbitrary limitations here

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/StarredFragment.java
johan12345: I agree, we should just share the JSON text instead or not offer the feature on Android <4.4 at all. Android 4.4 and above is already [91% of all Android devices](https://developer.android.com/about/dashboards/index.html).

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/StarredFragment.java
kedarmp: Alright I've handled JSON being split onto multiple lines.

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/StarredFragment.java
kedarmp: It has been done!

opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/StarredFragment.java
kedarmp: Sure :)

opacclient/opacapp/src/main/AndroidManifest.xml
kedarmp: Yep, the permission is unnecessary considering the fact that we do not need Android <4.4 devices to have a file in the first place.

PR #486, opened 2017-09-06
Implement export-import favorites list

kedarmp:
The date and the name of the commit may confuse you, but I have used git rebase. The new changes however, have been incorporated. I'm still learning git. Thanks!

### DIFF
--- a/opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/StarredFragment.java
+++ b/opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/StarredFragment.java
@@ -1,31 +1,33 @@
 /**
  * Copyright (C) 2013 by Raphael Michel under the MIT license:
- *
+ * <p>
  * Permission is hereby granted, free of charge, to any person obtaining a copy
- * of this software and associated documentation files (the "Software"), 
- * to deal in the Software without restriction, including without limitation 
- * the rights to use, copy, modify, merge, publish, distribute, sublicense, 
- * and/or sell copies of the Software, and to permit persons to whom the Software 
+ * of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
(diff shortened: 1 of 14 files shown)


raphaelm:
Thank you for this and so sorry for the late reply! :( Sadly, my next one will be late as well since I'm on vacation next week.

> The date and the name of the commit may confuse you, but I have used git rebase.

:+1: It is actually good practice to use rebase to make a PR look clean once its in a nearly-done stage :) For the next time: You don't even need to create a new pull request, you can just force-push to the new branch, GitHub will handle it.

> The new changes however, have been incorporated.

Yep, looks great! I tested it on both relevant Android versions and it works beautifully. I only have very little more suggestions:

* The icons are probably imperfect, since (a) no network is involved and the cloud symbol is misleading since it suggests synchronization and (b) it is unclear which one is up and which one is down. I like [these](https://materialdesignicons.com/icon/import) [two](https://materialdesignicons.com/icon/export) a little better but they also look like login/logout. I think in this case it might even be better do always hide the options into the overflow menu where they have text. What do you think? @johan12345, do you have an opinion?

* The reason Android Lint and Travis are complaining is that you renamed ``<string name="export_starred">Export</string>`` to ``<string name="share_starred">Share</string>``. You also need to do this rename in the translations, i.e. for example in ``values-de/strings.xml`` etc. for the other languages.

* Under old Android versions, if no custom picker is installed, "An error occured during import" is all that is given as an error message. For this case, do you think it'd be possible to give a more meaningful message?

* When saving the file, the default filename is just ``share.json``. I think it should be something like ``webopac_starred_<libname>.json`` such that users can find it again.

After I return from vacation, I'll merge this right away :)

raphaelm:
Hi @kedarmp! :) Everything alright or is there anything I can help you with?

johan12345:
As there was no response from @kedarmp , I have now implemented fixes for the points mentioned above. If @raphaelm agrees, we can merge this now.

PR #503, opened 2018-02-09
Add option to disable loading of covers on mobile data

### LINKED ISSUE #502: Add option to disable loading of covers on mobile data
**Hey!** If you ever wanted to contribute to an Open Source project, but didn't know how, you just found a perfect place to start. We try to write this issue as instructive as possible, but as we're humans we almost certainly forgot to write something important in here. If you want to try to tackle this issue and get stuck at _any_ point, please do not hesitate and ask us, either here on GitHub or via email at info@opacapp.de. We'll be very happy to help! 

We think this issue is very suitable to people who are just starting with Open Source or Android development in general and have never done it before. We'd like to ask people with wide experience in contributing to leave this issue to beginners. Thanks!
## How to get started

First of all, we recommend that you [install the app](https://play.google.com/store/apps/details?id=de.geeksfactory.opacclient) and play around with it so you know how it looks and feels. Next, you should use git to download this GitHub repository and open the folder `opacclient/` in Android Studio. Press the green "Play" button to run it on your device. If this works, you have a working development setup, yay! If it doesn't, feel free to ask us for help!

If you don't have an account in one of the [supported libraries](https://en.opacapp.net/libraries/), you can select the library that just has the name "Test". Note that this library is only included in the list if you compile the app yourself, you will not find it in the builds from the Play Store.
## What this is about

When you search for books in the catalog of a library, in many cases the app will display the cover picture of the book -- in the search result list, in the detail view or even both. A few users have requested that there should be an option in the app to disable downloading these covers while they are on a mobile data network so that the app uses slightly less data.
## How to do this

We recommend forking this repository and creating a special git branch for this issue before you start. This will make it easier for you to submit a pull request later. If you've never done this before, there are a lot of good guides out there and you can always ask us questions.
- Create a new setting (`CheckBoxPreference`) in the [settings.xml](https://github.com/opacapp/opacclient/blob/master/opacclient/opacapp/src/main/res/xml/settings.xml) file and add the corresponding strings in English. We can handle the translations for you. We think the setting best fits in the general category called "Web Opac". The default behavior should be to always load covers (as previously).
- We have recently decided to access all new preferences not directly, but using our [`PreferenceDataSource`](https://github.com/opacapp/opacclient/blob/master/opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/storage/PreferenceDataSource.java) class. So you should implement new methods there that access the new preference.
- Have a look at [`ResultsAdapter`, Line 156](https://github.com/opacapp/opacclient/blob/master/opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/ResultsAdapter.java#L156). As you can see there, if `item.getCover()`, which holds the cover URL, is set, the cover will be loaded. Here, you need to add an additional check the state of your newly-added setting and then, if needed, check whether the user is on a mobile network (e.g. using [`ConnectivityManagerCompat.isActiveNetworkMetered()`](https://developer.android.com/reference/android/support/v4/net/ConnectivityManagerCompat.html#isActiveNetworkMetered(android.net.ConnectivityManager)). If the cover should not be loaded, we can just show the media type icon as already implemented here.
- The same logic also needs to be implemented in [`SearchResultDetailFragment`, Line 1299](https://github.com/opacapp/opacclient/blob/master/opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/SearchResultDetailFragment.java#L1299) so that the `LoadCoverTask` is only executed when the preference is set to always load covers or when the device is on an unmetered network.
- Test your new setting on a device or emulator with a library that supports displaying covers.

We're excited to see your pull request! 

If you get stuck, don't hestitate to ask. We bound ourselves to a [Code of Conduct](https://github.com/opacapp/opacclient/blob/master/CODE_OF_CONDUCT.md) and we're committed to providing a welcoming experience and giving constructive, friendly and useful feedback on your contributions.

vivekscl:
Fixes #502 

As explained in the commit message:
```
There is no option to disable loading of covers while mobile data is switched on.

This is counterproductive for users who would prefer less data to be used when on a metered network.

As such, let's add an option to disable loading of covers while on mobile data.
```

### DIFF
--- a/opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/ResultsAdapter.java
+++ b/opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/ResultsAdapter.java
@@ -22,7 +22,9 @@
 package de.geeksfactory.opacclient.frontend;
 
 import android.content.Context;
+import android.net.ConnectivityManager;
 import android.support.annotation.DrawableRes;
+import android.support.v4.net.ConnectivityManagerCompat;
 import android.text.Html;
 import android.view.LayoutInflater;
 import android.view.View;
@@ -44,6 +46,7 @@ import de.geeksfactory.opacclient.networking.CoverDownloadTask;
 import de.geeksfactory.opacclient.objects.CoverHolder;
 import de.geeksfactory.opacclient.objects.SearchResult;
 import de.geeksfactory.opacclient.objects.SearchResult.MediaType;
+import de.geeksfactory.opacclient.storage.PreferenceDataSource;
 import de.geeksfactory.opacclient.utils.BitmapUtils;
 
 public class ResultsAdapter extends ArrayAdapter<SearchResult> {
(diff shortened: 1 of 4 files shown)


johan12345:
Thank you very much!

PR #534, opened 2018-10-21
Fix icon misaligned in landscape mode

foxtrot95:
Fixed icon size in relation to the space left and adjusted the top drawer height. No overlapping of the icon should occur in landscape mode anymore.


johan12345:
Looks good, thank you very much!

PR #535, opened 2018-10-22
Make status indicator circles accessible for color-blind people

foxtrot95:
Changed the status icons to make it more visible for color blind people. Adjusted also the size of the status indicators. The same colors were used as the one before.


johan12345:
Looks great, thank you!

PR #539, opened 2018-11-06
Improve lent item list format and add branch

StefRe:
Make items of lent items list look a bit nicer and clearer:
- Remove trailing delimiter for items without author (author string is empty but not null) (see second item in example below)
- Change title/author delimiter from ‘,’ to ‘.’ and print title in bold for better readability and clarity (see e.g. third item where is wasn't particularly clear where the title ends and the author starts)
- Remove details about person after ¬ sign (at least Dresden municipal library often has further information such as author, editor, producer etc. after the person's name) (see first and last item)

Add branch information to the list. I regularly borrow books from different branches of the same library. When I need to return a book I want to see what other books I hold from the same branch. This was only possible by clicking for details of each individual item which is quite cumbersome for a longer list.

comparison

### DIFF
--- a/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/objects/AccountItem.java
+++ b/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/objects/AccountItem.java
@@ -102,7 +102,7 @@ public abstract class AccountItem implements Serializable {
 
     /**
      * @return status of an item. Some libraries use codes like "E" for "first lending period", "1"
-     * for "lending period extended once", etc. Optional.
+     * for "lending period extended once" etc. or strings like e.g. "2x verl., 1x reserv.".  Optional.
      */
     public String getStatus() {
         return status;
@@ -110,7 +110,7 @@ public abstract class AccountItem implements Serializable {
 
     /**
      * Set status of an item. Some libraries use codes like "E" for "first lending period", "1" for
-     * "lending period extended once", etc. Optional.
+     * "lending period extended once" etc. or strings like e.g. "2x verl., 1x reserv.". Optional.
      */
     public void setStatus(String status) {
         this.status = status;
--- a/opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/adapter/AccountAdapter.java
+++ b/opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/frontend/adapter/AccountAdapter.java
@@ -71,6 +71,7 @@ public abstract class AccountAdapter<I extends AccountItem, VH extends AccountAd
         protected Context context;
         public TextView tvTitleAndAuthor;
         protected TextView tvStatus;
+        protected TextView tvBranch;
         protected View vStatusColor;
         protected ImageButton ivProlong;
         protected ImageButton ivDownload;
@@ -88,6 +89,7 @@ public abstract class AccountAdapter<I extends AccountItem, VH extends AccountAd
             this.context = itemView.getContext();
             tvTitleAndAuthor = (TextView) itemView.findViewById(R.id.tvTitleAndAuthor);
             tvStatus = (TextView) itemView.findViewById(R.id.tvStatus);
+            tvBranch = (TextView) itemView.findViewById(R.id.tvBranch);
             vStatusColor = itemView.findViewById(R.id.vStatusColor);
             ivProlong = (ImageButton) itemView.findViewById(R.id.ivProlong);
             ivDownload = (ImageButton) itemView.findViewById(R.id.ivDownload);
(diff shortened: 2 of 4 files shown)


johan12345:
Thank you very much! The first two improvements look great.

Regarding the ¬ sign: It might be that other libraries or other OPAC systems use that sign for different purposes. For example, some use it in the title of an item to mark that it starts with an article (Der/die/das/ein/eine, etc.), e.g. "¬Das¬ Herz der Ozeane". I currently don't have an example for other uses in the author field, but still, it might be better to implement this cleanup in the respective libopac API implementation instead of doing this directly in the account view for all libraries. @raphaelm , what do you think about this?

Adding the branch to the displayed data is also a good idea, but we have to test how it looks like if the data shown in the Status field is longer than just a date.

StefRe:
My understanding from the [code](https://github.com/opacapp/opacclient/blob/505b45bfd12ddaeacc7cf18081506f9690ec09a5/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/objects/AccountItem.java#L104) was that the status information after the deadline is just one letter, so I split the row 50 : 50 to give enough room. If there might occur longer status entries than we could change this ratio to say 60 : 40, but I'm not sure if the remaining length of the ellipsized branch name is sufficient to discern different branches (in Dresden, it's no problem as they all start with 'Bib. ' and then comes the branch location, but if the first word is for instance 'Zweigstelle ...' then this could become a problem). 
On the other hand, one could simply append the branch to the status field (after a couple of spaces) but i thought it would look much nicer if the branch fields of all items were neatly aligned at the same position.

raphaelm:
> As for the ¬ sign: I would keep the full (unprocessed) author field in the details view (and hence don't do any cleanup in the API) and do the cleanup in the lent items list just for optical reasons. If by error we remove some important piece of information here (and who knows what it's used for in all cases), it'll still be available in the details view.

Johan is right: In library databases it is often used to denote parts of a title that should be ignored when sorting a list, so just discarding after the first occurence will often discard the full string.

I will comment on the other issues in the next days.

StefRe:
> ...denote parts of a **title** ...

I'm only talking about the author field, not the title. You're right for the title - this would leave us with many empty titles as the article is usually the first word.
For authors this situation is quite different (at least with Dresden municipal library) as you can see from the followoing examples:
```
Alemann, Ulrich ¬von ¬[Verfasser]
Balzac, Honoré ¬de
Beethoven, Ludwig ¬van ¬[Komponist]
Bentheim und Steinfurt, Elna-Margret ¬zu ¬[Verfasser]
Kwast, Ernest ¬van ¬der ¬[Verfasser]
Meiklokjes, Enie ¬van ¬de
```
As we don't know whether all libraries follow that scheme let's do the cutting not at "¬" but at "¬[" - a combination that for sure will only be encountered **after** the main word(s).

Additionally we could then just delete all the remaining ¬ signs in both the author and title fields.

raphaelm:
> I'm only talking about the author field, not the title. You're right for the title - this would leave us with many empty titles as the article is usually the first word.

Sorry, missed that. Still, by the same logic I wouldn't be surprised to see "¬Dr. Einstein, Albert" in some systems. But you're right, it wouldn't be too bad to lose the author field. Deleting after ``¬[`` seems save, though.

StefRe:
I modified the PR to cut off the author field after `¬[` instead of `¬`.

raphaelm:
Looked over the code and screenshots, I like it.

The only think left to think about is the status field length, as Johan said. @johan12345 Do you happen to know what kinds of content it can contain?

johan12345:
The status field can definitely contain more than just a single-letter code, we should update the Javadoc for that.. Two examples would be "Noch nicht verfügbar" (WinBiap, reserved items) or "2x verl., 1x reserv." (IOpac, lent items).
Maybe we need to optionally move the branch to a third row when the status and branch are too long to fit on one line together? This would also not only apply with very long status texts, but probably also on some narrower phones.

Also, another thing: As far as I see, the new `tvBranch` field will also be shown for reserved items (in `ReservationsAdapter`), but not used in any way. @StefRe , could you either hide it there or also show the corresponding branch of the reservation?

johan12345:
> If you like this solution I'll amend the PR accordingly.

Looks good, thank you!

> which one is the branch for the reservation: home branch or lending branch? Couldn't figure it out from the code.

For reservations, there is only one branch:
https://github.com/opacapp/opacclient/blob/505b45bfd12ddaeacc7cf18081506f9690ec09a5/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/objects/ReservedItem.java#L68

For lent items, there are two branches:
https://github.com/opacapp/opacclient/blob/505b45bfd12ddaeacc7cf18081506f9690ec09a5/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/objects/LentItem.java#L59
https://github.com/opacapp/opacclient/blob/505b45bfd12ddaeacc7cf18081506f9690ec09a5/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/objects/LentItem.java#L73
where homeBranch is the branch the item belongs to and lending branch is the branch it is lent from. But there are many libraries that only show one branch, and then we usually put it into the homeBranch field AFAIK.

StefRe:
I changed the PR as discussed (provided enough room for longer status field, hid branch at reservations, updated Javadoc) and reduced the gap between the data and the buttons for a more balanced visual impression.

raphaelm:
Thank you so much!

PR #545, opened 2018-12-18
Fix typo

StefRe:


opacclient/opacapp/src/main/res/values/strings.xml
```diff
@@ -281,7 +281,7 @@
     <string name="notification_repeat">Repeat notifications</string>
     <string name="update_info_title">Happy Birthday Opac App!</string>
     <string name="update_info_text">This May, our app turned five years old. We\'re celebrating the birthday with a new logo and want to thank all of you for all the feedback that we received.</string>
-    <string name="update_info_text_second">If you want to help us keep the app sustainable in the future, talk to your library: By buying the Plus Edition of the app, they allows us to continue developing this app and at the same time unlocks a set of useful features for you. Check our website www.opacapp.net for more details!</string>
+    <string name="update_info_text_second">If you want to help us keep the app sustainable in the future, talk to your library: Buying the Plus Edition of the app allows us to further develop it and at the same time unlocks a set of useful features for you. Check our website www.opacapp.net for more details!</string>
```
johan12345: Hm, I see that there is a typo (*allows* should be *allow*, or even better, *enable*), but the way that you changed it might suggest that the Plus Edition can be bought by the users themselves. Instead, the Plus Edition needs to be bought by the library (-> that's why the *they* is there).

StefRe:
OK, replacing _allow_ by _enable_ makes it possible to keep the _they_.  But then there's still a problem with the second half of the sentence: either change _unlocks_ to _unlock_ (if they do it by buying the app) or insert a _this_: ... and at the same time _this_ unlocks a set ... . With English not being my native language I'm not sure which one is better, they both sound a bit clumsy to me. 

Should we then change it to `By buying the Plus Edition of the app, they enable us to continue developing this app and at the same time this unlocks a set of useful features for you.`
Shall I change the PR or can you edit it directly?

johan12345:
> Should we then change it to `By buying the Plus Edition of the app, they enable us to continue developing this app and at the same time this unlocks a set of useful features for you.`
> Shall I change the PR or can you edit it directly?

Yes, this sounds good! The only thing I would add is a comma before `at the same time` (also not a native English speaker, though). I will add it directly.

PR #548, opened 2018-12-22
Singular forms for expired items notification

StefRe:
Just a small beautification.
Add singular forms to avoid texts like "1 items have expired". Change de, es, fr, it, ja and pl translations accordingly. Change and complete cs translation accordingly. Change incomplete translations for el and tr (to the best of my knowledge, as I don't know these languages).
out


johan12345:
Looks good to me, thank you! (although I'm also not qualified to check all the translations ;) )

PR #550, opened 2018-12-29
Get media types for lent items in SISIS api

StefRe:
I missed the media type information in the lent items details because sometimes this is good to have, especially if you have lent different media types on a certain subject with more or less the same titles. Having fetched the media types, they're now available not only in the details view, but also show up in the lent items and reservations lists where they take up quite some space (especially on smaller displays). That's why I also added an option to show or hide the media type icons in the lent items and reservations lists (default is 'show'). There are a lot of libraries using SISIS in Germany and I wasn't sure if all users would find it nice to suddenly see the media type items in the list.
Getting the media types for lent items turned out to be very similar to getting them for search results, so I just reused the existing code by putting it into a function. My understanding of the api doc was that the `format` field is just a fallback option if `mediatype` can't be determined, so I only populate this field if no mediatype can be derived.  The new `MediaTypeOrFormat` inner class is only needed to let `getMediaTypeOrFormat` return two values of different types.

### DIFF
--- a/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SISIS.java
+++ b/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SISIS.java
@@ -448,39 +448,12 @@ public class SISIS extends OkHttpBaseApi implements OpacApi {
         List<SearchResult> results = new ArrayList<>();
         List<CompletableFuture<Void>> futures = new ArrayList<>();
         for (int i = 0; i < table.size(); i++) {
-            Element tr = table.get(i);
             SearchResult sr = new SearchResult();
-            if (tr.select("td img[title]").size() > 0) {
-                String title = tr.select("td img").get(0).attr("title");
-                String[] fparts = tr.select("td img").get(0).attr("src")
-                                    .split("/");
-                String fname = fparts[fparts.length - 1];
-                MediaType default_by_fname = defaulttypes.get(fname
-                        .toLowerCase(Locale.GERMAN).replace(".jpg", "")
-                        .replace(".gif", "").replace(".png", ""));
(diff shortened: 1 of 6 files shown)


raphaelm:
Hmm, I'm not sure if we actually need a setting for this. Is there a good reason not to want this, @johan12345?

johan12345:
I'm also against adding an additional setting for this - too many settings are also confusing. If you think they take up too much space on smaller displays, we could have a look whether we can make them smaller there.

Yes, `format` is basically a fallback if we can't determine the `mediatype` - previously we only had `format` in the account view, and some APIs don't yet support `mediatype` (or it's not easily possible to determine the `mediatype` in the account view because they use a different way to show the media type there than in the search results).

StefRe:
I don't necessarily need the setting for hiding the media types, it looks OK on my device. I just remembered we had a discussion about limited space on smaller displays in #539, that's why I kind of preemptively introduced this option to give the user the possibility the restore the familiar view if they find the icons too intrusive and hence the remaining text space too small.
Making the icon smaller is not a option in my opinion, as the same space may be used by cover images and, although already a bit larger than the icons, they are rather small and often hardly discernible on a phone.
So I tried to further reduce the margins/padding:
* Original view:
   original  

* New view c24c18dcd3600b8ae136cdbd158ae7645b8beb3a: 
   new  

* More compact new view:
reduced

I like the last one best (the 4dp color indicator field on the very right margin is white here, when it's colored we get an even more balanced visual impression). If you like this I'll push it as a third commit and you just don't merge the second commit 83cbd25f9faa813acddca4553330618cc1e4f64c with the option settings.

raphaelm:
I think it looks good enough and not to crowded, so yes, feel free to push the third commit. We can omit the second while merging :)

StefRe:
pushed third commit

PR #554, opened 2019-01-08
Shrink icon size in search results to 40 dp

StefRe:
This ensures **consistency in display between lent items list and search results list**. In the former (listitem_account_item.xml), there are individual image views for cover images and icons of size 56 and 40 dp, respectively, which are shown/hidden as needed. In the search results list, both covers and icons are displayed using the same 56 dp image view, which results in disproportionately big icons as compared to cover images. This is especially striking when book icons (landscape format) are mixed with cover images (usually portrait format).
Instead of introducing separate image views for covers and icons as in the lent items list, we just set a padding of 8 dp in case of icons being displayed in the ivType image view to shrink them to 40 dp. As the conversion from dp to px is a relatively expensive operation and needs a context, the calculated px value is cached.

|OLD | NEW |
|---|---|
|_icon56  | _icon40 |

(Sorry for being a bit pedantic about the views, but as we say in German _auch das Auge ißt mit_)


johan12345:
Looks good, thanks!

PR #563, opened 2019-04-30
Make HttpClientFactory work in non-Android contexts

sebkur:
I'm trying out using `libopac` in non-Android contexts, i.e. a normal Java project. I'm facing the problem that access to file `ssl_trust_store.bks` in implemented in a way that works on Android only because it assumes files to be available at `res/`. You can see what I'm doing here in [my fork of the sample app](https://github.com/sebkur/libopac-sample-gradle). The problem arises not with the default config of the sample (Bremen), but for the Berlin config.

I worked around this limitation by moving the file `ssl_trust_store.bks` to `src/main/resources` of module `libopac` and then loading it like this:

    Thread.currentThread().getContextClassLoader().getResourceAsStream(ssl_store_path);

That should work on Android, too. I do that in production code in apps very often. It would be good if you could double-check anyway.


johan12345:
Hi,

thanks for the contribution! I think this change looks reasonable, but I haven't tested it yet. I will be travelling for three weeks starting next weekend and need to implement some urgent fixes for the OPAC app before that, so I might not get around to this. Maybe @raphaelm would have time to take a look in the next weeks?

raphaelm:
Hi @sebkur, sorry it took me so long to take a look at this.

I haven't checked whether it actually works on Android with getResourceAsStream, but I assume it does.

Is there a reason you copied the file instead of moving it? I would want to avoid having it twice, this will only cause confusion. Also, there are some references to the file in ``tools/``, can you update them as well?

sebkur:
Thanks for having a look!

Moving the file makes sense to me, too. I updated the commit accordingly and also changed the references in `tools/` as suggested.

raphaelm:
As you can see on Travis, this is currently breaking the Android build because of this:

```
/home/travis/build/opacapp/opacclient/opacclient/opacapp/src/main/java/de/geeksfactory/opacclient/networking/AndroidHttpClientFactory.java:48: error: cannot find symbol
                R.raw.ssl_trust_store);
```

Can you change that to the new system as well?

sebkur:
Okay, didn't see that, but fixed it now.

Travis still fails, but this seems to be an unrelated issue that also occurs in the current master branch.

sebkur:
Just rebased to the current master and now Travis doesn't complain anymore

johan12345:
Looks good to me. Thanks and sorry that it took so long!

sebkur:
Great, thanks. It would be really cool if you published a version 6.0.4-beta2 to bintray :)

raphaelm:
Done! :)

sebkur:
Thanks!

PR #568, opened 2019-11-21
replace evernote/android-job with WorkManager

johan12345:
[android-job](https://github.com/evernote/android-job) is deprecated and uses [GcmNetworkManager](https://developers.google.com/android/reference/com/google/android/gms/gcm/GcmNetworkManager), which will also be deprecated next year. Thus, we replace it with the new [WorkManager](https://developer.android.com/topic/libraries/architecture/workmanager) library.


johan12345:
tests are fixed now and the job seems to run correctly

PR #571, opened 2020-01-08
Add missing setVolumes and fix typos

StefRe:
so that volumes has a setter and getter for the whole list the same way as copies has.

I came across the missing setter in `slub.kt` where I get the volumes list by mapping `Volumes()` onto a JSON array and it's therefore more idiomatic to set the whole list I get this way as opposed to adding all the elements to the list.


johan12345:
looks good, thanks!

Issue #577, opened 2020-02-23
Secure Connection to the library server failed for Ettlingen


dentes-purgo:
See attached Image:
![grafik](https://user-images.githubusercontent.com/61378839/75111291-8c496900-5638-11ea-9387-87002cc1a941.png)
Yes, i know that Ettlingen has no support contract (see screenshot, ettlingen.json). Maybe I will try to fix it myself.

dentes-purgo:
It's the following error:  

> java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.

thrown at 

> RealConnection.java:319 => sslSocket.startHandshake();  
i think ...

maybe this can be helpful...

dentes-purgo:
Seems to be a generall problem concerning  all librarys listed here: https://web-opac.kivbf.de/ (tried some, all terminating with the same error)

dentes-purgo:
seems that they have a new certificate which is not trusted by the app 
comes from Sectigo limited
which itself is signed by  USERTrust RSA Certification Authority

johan12345:
Then the certificate authority is probably not included in Android's default trust store (it might still work in e.g. the Chrome for Android browser, but that is updated separately).

We have [a script](https://github.com/opacapp/opacclient/blob/master/tools/add_certificate.bash) for downloading a certificate and adding it to the app's own trust store - that should help.

dentes-purgo:
Sure. But that isn't something a newbe like me is supposed to do, isn't it :-) ?

johan12345:
Why not? It's running one bash command, testing if it works, and sending a Pull request.

dentes-purgo:
Sorry. I deleted my comment by mistake.  And I needed a little bit of time to grasp that I have to ommit the https:// in the name of the server.
And it would have been helpful if I had known that I have to run it from the opacclient directory and not from the tools directory to prevent me from getting

> ClassNotFoundException: org.bouncycastle.jce.provider.BouncyCastleProvider

Now I added it and the app seems to work again in the emulator.

johan12345:
That sounds good! Sorry that there is not too much documentation about these things. But keeping the app working for thousands of libraries not a trivial task, and that's exactly why we have [switched to the support contract model](https://opac.app/de/support-policy/) after all...

If your updated keystore works, we would gladly accept this change as a [Pull Request](https://github.com/opacapp/opacclient/compare) and include it into the next app release.

dentes-purgo:
[...]

PR #578, opened 2020-02-23
#577 Secure Connection to the library server failed for Ettlingen

dentes-purgo:
Added the certificate for https://web-opac.kivbf.de/ (because they have a new one since 02/18/2020


johan12345:
Thanks!

PR #581, opened 2020-06-30
A bunch of miscellaneous upgrades

TacoTheDank:
Sorry this is so big, didn't mean for that.

### 1st commit
- Delete some unnecessary files
- Clean up duplicates in .gitignore

### 2nd commit
- Update gradle wrapper to 6.5 using git bash

### 3rd commit
- Reformat and rearrange gradles to be easier to read
- Update Android Studio 3.6.1 -> 4.0.0 ([changelog](https://developer.android.com/studio/releases))
- Update bintray plugin 2.8.4 -> 2.8.5 ([changelog](https://bintray.com/jfrog/jfrog-jars/gradle-bintray-plugin#release))
- Update coveralls plugin 2.4.0 -> 2.8.3 ([changelog](https://github.com/kt3k/coveralls-gradle-plugin#release-history))
- Grouped dependencies for easier reading

### 4th commit
- libopac:
  - Updated Jsoup 1.8.3 -> 1.13.1 ([changelog](https://github.com/jhy/jsoup/blob/master/CHANGES))
  - Updated `org.apache.httpcomponents:httpmime` 4.5.2 -> 4.5.12 ([changelog](https://archive.apache.org/dist/httpcomponents/httpclient/RELEASE_NOTES-4.5.x.txt))
  - Updated joda-time 2.9.9 -> 2.10.6 ([changelog](https://www.joda.org/joda-time/changes-report.html))
  - Updated OkHttp3 versions -> 3.12.12 ([changelog](https://square.github.io/okhttp/changelog_3x/))
  - Updated streamsupport-cfuture 1.6.0 -> 1.7.2 ([changelog](https://github.com/stefan-zobel/streamsupport/releases))
  - Updated JUnit 4.12 -> 4.13 ([changelog](https://github.com/junit-team/junit4/blob/HEAD/doc/ReleaseNotes4.13.md))
    - Fixed deprecations in tests
  - Updated hamcrest-library 1.3 -> 2.2 ([changelog](https://github.com/hamcrest/JavaHamcrest/releases))
    - Fixed deprecations in tests

- meaningdetector:
  - Updated `org.bouncycastle:bcprov-jdk15on` 1.51 -> 1.65.01 ([changelog](https://www.bouncycastle.org/releasenotes.html))

- opacapp:
  - Updated AndroidX Multidex 2.0.0 -> 2.0.1 ([changelog](https://developer.android.com/jetpack/androidx/releases/multidex))
[...]

PR #584, opened 2020-07-13
[WIP] Jsoup upgrade

johan12345:
@TacoTheDank already tried to upgrade JSoup in #581, but this lead to some test failures in libopac. We need to investigate these and also do some testing of search in different OPAC systems to make sure that we don't introduce bugs with the JSoup upgrade.

### DIFF
--- a/opacclient/libopac/build.gradle
+++ b/opacclient/libopac/build.gradle
@@ -17,7 +17,7 @@ dependencies {
 // Other
     // Use older org.json version to have an environment equivalent to the android platform
     api 'org.json:json:20090211'
-    implementation 'org.jsoup:jsoup:1.8.3'
+    implementation 'org.jsoup:jsoup:1.13.1'
     implementation 'org.apache.httpcomponents:httpmime:4.5.12'
     api 'joda-time:joda-time:2.10.6'
     api 'com.squareup.okhttp3:okhttp:3.12.12'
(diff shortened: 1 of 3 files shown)


johan12345:
Since version 1.11.1, Jsoup [replaces &nbsp; characters with spaces](https://github.com/jhy/jsoup/commit/d3f02400eca1c5cbf84212c46cebcff511536655). This was the cause of two of the test failures, and will probably also affect some implementations in the other APIs. We should further investigate this.

johan12345:
The test failures should be fixed now, but more testing of the parts of libopac not covered by the automated tests is required (see above).

johan12345:
@raphaelm You updated JSoup in 11833f3234, have you tested if this doesn't break anything? Especially the `&nbsp;` replacement mentioned above may be important for some of the APIs.

At least the other changes in this PR probably need to be merged to fix the tests.

johan12345:
see also #639 for other things broken by the dependency upgrade

raphaelm:
ouch, sorry, somehow missed this. I think I needed to upgrade to replace the (no longer really working) stetho with flipper for debugging. I'll look into this and downgrade or fix, just need to find the time.

johan12345:
Ah, okay. Regarding Stetho issues: I think a recent Chrome update broke it, also noticed that in https://github.com/johan12345/EVMap. It still works in other Chromium-based browsers though (Microsoft Edge, Brave). These the corresponding Stetho and Chromium issues: https://github.com/facebook/stetho/issues/696, https://bugs.chromium.org/p/chromium/issues/detail?id=1187142

raphaelm:
I reverted for now

PR #585, opened 2020-07-14
Adjust codestyle to an .editorconfig

TacoTheDank:
This will make the codestyle consistent throughout the project.

(This removes unused imports as a side effect, but that should be fine.)

No actual code changes were made, as this was done automatically by Android Studio.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+# editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true


johan12345:
Hm, adding an `.editorconfig` is probably good, but I'm not sure if we want to reformat all existing code, because e.g. it will break `git blame`.

Most of the existing code should already have been formatted using the Android Studio defaults (except some very old code from before Android Studio existed).

Also, it would take ages to manually review all the changes in the PR to verify that no actual code changes were made (this would be a good place to sneak in malware). So *if* we reformat all existing code, we will do it ourselves.

TacoTheDank:
@johan12345 Good point, I'll add just the .editorconfig then.

johan12345:
Thanks!

PR #587, opened 2020-07-18
Fix: No signature of method: ...cleanDirectory()

asdoi:
Everytime I run `./gradlew downloadJson` I get the following error:
No signature of method: static org.gradle.util.GFileUtils.cleanDirectory()


johan12345:
Right, it looks like the method was [removed](https://github.com/gradle/gradle/commit/1c07bebdc54b57fb767fb283557c015a2fa85e41) recently, and so this error occurs since Gradle was upgraded in #581.

It seems that the org.gradle.util package is [not considered public API](https://docs.gradle.org/current/userguide/authoring_maintainable_build_scripts.html#sec:avoiding_gradle_internal_apis) - so maybe we should rather avoid using it altogether and replace it e.g. with Apache Commons IO's `org.apache.commons.io.FileUtils.cleanDirectory()`.

johan12345:
I implemented this in 8036bd957b524a8bc388395808d93409ebcb42d9.

Issue #582, opened 2020-07-05
Secure connection to Stadtbibliothek Würzburg failed


dueringa:
See also issue #577 

Same issue as in this ticket.

Stadtbibliothek Würzburg uses the same Root Certificate, so I thought the linked issue fixed this.

I added the root certificate as mentioned in the linked issue anyway. The outut said:

```
$ LC_ALL=C ./tools/add_certificate.bash opac.stadt.wuerzburg.de
Adding certificate to opacclient/libopac/src/main/resources/ssl_trust_store.bks...
Certificate already exists in system-wide CA keystore under alias <debian:usertrust_rsa_certification_authority.pem>
Do you still want to add it to your own keystore? [no]:  yes
Certificate was added to keystore
[Storing opacclient/libopac/src/main/resources/ssl_trust_store.bks]
```

 However, I'm still getting the error message.

I deleted all app data and re-added the library, didn't change the behavior.

adb logcat says

```
07-05 20:38:50.345 13387 13424 W System.err: javax.net.ssl.SSLHandshakeException: java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.
07-05 20:38:50.348 13387 13424 W System.err: 	at com.android.org.conscrypt.ConscryptFileDescriptorSocket.startHandshake(ConscryptFileDescriptorSocket.java:229)
07-05 20:38:50.348 13387 13424 W System.err: 	at okhttp3.internal.connection.RealConnection.connectTls(RealConnection.java:11)
07-05 20:38:50.348 13387 13424 W System.err: 	at okhttp3.internal.connection.RealConnection.establishProtocol(RealConnection.java:9)
07-05 20:38:50.348 13387 13424 W System.err: 	at okhttp3.internal.connection.RealConnection.connect(RealConnection.java:15)
07-05 20:38:50.348 13387 13424 W System.err: 	at okhttp3.internal.connection.StreamAllocation.findConnection(StreamAllocation.java:36)
07-05 20:38:50.348 13387 13424 W System.err: 	at okhttp3.internal.connection.StreamAllocation.findHealthyConnection(StreamAllocation.java:1)
07-05 20:38:50.348 13387 13424 W System.err: 	at okhttp3.internal.connection.StreamAllocation.newStream(StreamAllocation.java:6)
07-05 20:38:50.351 13387 13424 W System.err: 	at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:5)
[...]

PR #593, opened 2020-09-08
Add intermediate certificate to trust store for library Wuerzburg

dueringa:
Their webserver config doesn't send the intermediate certificate, relying on download by browser (which doesn't work with apps). I contacted the library, they won't fix it.

Added certificate with alias `opac.stadt.wuerzburg.de-65ff7287`

See discussion in #582


johan12345:
Thanks!

PR #594, opened 2020-10-14
Fix array index range of JSONArray extension functions

StefRe:
The rangeTo operator ".." includes the end element which leads to an array index
out of range error. The "until" function does not include the end element.
see https://kotlinlang.org/docs/reference/ranges.html


johan12345:
thanks!

Issue #595, opened 2020-10-15
SLUB API - pending items


StefRe:
The API works (catalog and account features) but there are still some TODOs to be completed from #553:

- [ ]  [get status](https://github.com/opacapp/opacclient/blob/27395799f8318c476a72162f2f1ee8ba2c7325a5/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SLUB.kt#L152): low priority due to the limited usefulness of the status as in many many cases it just says "click for more information about availability" (in the online web catalogue), so you'll have to open the details page anyway.  
Implementation idea: use [asyncGet](https://github.com/opacapp/opacclient/blob/c89e0c42d56585cff0667cee9e339c329906cb9d/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/OkHttpBaseApi.java#L237) and then put all the CompletableFutures constructed into a list before calling get() on them (thanks to @johan12345 [here](https://github.com/opacapp/opacclient/pull/553#discussion_r502957405) and [here](https://github.com/opacapp/opacclient/pull/553#issuecomment-574281016))

- [ ] [references](https://github.com/opacapp/opacclient/blob/27395799f8318c476a72162f2f1ee8ba2c7325a5/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SLUB.kt#L246): libero links must be resolved to `rsn `entries in current SLUB catalog and they then in turn to `id `entries (similar to next below)

- [ ] [`id` of reserved items](https://github.com/opacapp/opacclient/blob/27395799f8318c476a72162f2f1ee8ba2c7325a5/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SLUB.kt#L423) `bc` entries to be resolved to `id` entries by intercepting redirection

- [ ] [`id` of lent items](https://github.com/opacapp/opacclient/blob/27395799f8318c476a72162f2f1ee8ba2c7325a5/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SLUB.kt#L472) `bc` entries to be resolved to `id` entries by intercepting redirection (same as above)

- [ ] [actual ready date of stack request](https://github.com/opacapp/opacclient/blob/27395799f8318c476a72162f2f1ee8ba2c7325a5/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SLUB.kt#L425): works but could be refined, need more examples to examine

[...]

PR #598, opened 2020-10-23
SLUB: Get Results by Id for Lent and Reserved Items

StefRe:
This resolves items 3 and 4 of #595.

### DIFF
--- a/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SLUB.kt
+++ b/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SLUB.kt
@@ -159,15 +159,22 @@ open class SLUB : OkHttpBaseApi() {
 
     override fun getResultById(id: String, homebranch: String?): DetailedItem {
         val json: JSONObject
-        val url =
-                if (id.startsWith("id")) {
-                    "$baseurl/$id/"
-                } else {
-                    val noRedirectClient = http_client.newBuilder()
-                            .followRedirects(false)
-                            .build()
-                    httpHead("$baseurl/$id/", "Location", "", noRedirectClient)
-                } + "?type=1369315142&tx_find_find[format]=data&tx_find_find[data-format]=app"
+        val url = when {
+            id.startsWith("id/") ->
+                "$baseurl/$id/"
+            id.startsWith("bc/") || id.startsWith("rsn/") ->
+                http_client.newBuilder()
+                        .followRedirects(false)
+                        .build()
+                        .run {
+                            httpHead("$baseurl/$id/",
+                                    "Location",
+                                    "",
+                                    this)
+                        }
+            else -> // legacy case: id identifier without prefix
+                "$baseurl/id/$id/"
+        } + "?type=1369315142&tx_find_find[format]=data&tx_find_find[data-format]=app"
         try {
             json = JSONObject(httpGet(url, ENCODING))
         } catch (e: JSONException) {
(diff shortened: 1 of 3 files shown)


opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SLUB.kt
```diff
@@ -159,19 +159,26 @@ open class SLUB : OkHttpBaseApi() {
 
     override fun getResultById(id: String, homebranch: String?): DetailedItem {
         val json: JSONObject
+        val url =
+                if (id.startsWith("id")) {
+                    "$baseurl/$id/"
+                } else {
+                    val noRedirectClient = http_client.newBuilder()
```
johan12345, line 166: Do you really need to do this separate HEAD request to capture the redirected URL? You could also simply to the GET request (which should follow the redirect) and then use that as your JSON body directly, right?

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SLUB.kt
```diff
@@ -159,19 +159,26 @@ open class SLUB : OkHttpBaseApi() {
 
     override fun getResultById(id: String, homebranch: String?): DetailedItem {
         val json: JSONObject
+        val url =
+                if (id.startsWith("id")) {
+                    "$baseurl/$id/"
+                } else {
```
johan12345, line 165: Maybe you should also check here whether the ID starts with `bc/` and provide a fallback otherwise (e.g. for items saved on the favorites list before this change).

johan12345 (changes requested):
Looks good to me, except the two comments added above. Thanks!

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SLUB.kt
StefRe: Yes, the redirect leads to the web catalog view, i.e. it returns a html, not a json. In order to get the json for the redirected site you need to add `?type=1369315142&tx_find_find[format]=data&tx_find_find[data-format]=app`. 
Of course all this could be done using an interceptor, but I found it easier to implement and test it this way.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SLUB.kt
StefRe: Hm, old items from the favorites list seem to be the only possible case where you could encounter an id without a prefix. I think as the API is brand new no-one would have already used it with favorites.  
There are also `rsn` identifiers (not used yet) than can be resolved to `id` identifiers the same way with this code, so the correct and universal way to verify whether the identifier is an old un-prefixed `id` identifier would be by checking that neither the third nor the fourth character is a `/`. If you think we really need it I can add it, no problem - it's your choice.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SLUB.kt
johan12345: Ah, okay, yes, if the URL of the redirect needs to be changed that is probably the easiest solution.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SLUB.kt
johan12345: Hm, I would still try to be as explicit as possible (also because the code then better documents which possibilities there are for an ID) - so check for all cases, such as:
```kotlin
val url = when (id) {
    startsWith("id/") -> "$baseurl/$id/"
    startsWith("bc/"), startsWith("rsn/") -> // ...
    else -> "$baseurl/id/$id"
}
```

StefRe: done, see 360b9e5 (though the `when` didn't work out so nicely as suggested).

johan12345: Ah, right, I thought `when` would work like this, but apparently it doesn't. Still okay though :)

johan12345 (approved):
Thanks!

PR #599, opened 2020-10-29
SISIS: fix regex for cover URL for Dresden Städtische Bibliotheken

StefRe:
In Dresden, they changed the URL in the AJAX section: it now starts with "/webOPACClient/", which is also part of the opac_url.
OLD: `"jsp/result..." ` (used for instance in Erfurt)
NEW: `"/webOPACClient/jsp/result..."` (used in Dresden)

So we retrieve just the part starting with `"jsp/result..."`

Eliminate duplicate definition of the regex in `parse_search` and `loadDetail.`

### DIFF
--- a/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SISIS.java
+++ b/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SISIS.java
@@ -196,8 +196,6 @@ public class SISIS extends OkHttpBaseApi implements OpacApi {
     protected long logged_in;
     protected Account logged_in_as;
     protected static final String ENCODING = "UTF-8";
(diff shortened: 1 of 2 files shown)


johan12345:
The new kind `"/webOPACClient/jsp/result..."` seems to be a root-relative URL. Instead of solving this within the regex, I would rather suggest actually resolving the URL (so this is not dependent on the root path being called `webOPACClient`):
```java
String path = matcher.group(1);
String url = new URI(opac_url + "/").resolve(path).toString();
```
Does this work?

StefRe:
Yes, it works.  
As URI methods may throw exceptions I refactored it into a separate method. This makes unit testing it much easier too.

johan12345:
That looks good, thank you!

PR #603, opened 2020-11-30
SLUB: add some more media types

StefRe:
There is no specific media type for `"Norm"` (norms or standards). As these are mostly electronic resources (at least the current norms) I mapped them to the media type `EDOC`. Or do we need a new, more specific mediatype for it?


johan12345:
I think it's fine like this :)

PR #606, opened 2020-12-11
Show status text for copies in color

StefRe:
This is to mimic the SLUB web catalog behavior (left web catalog, right app with this PR):
  &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;      

This also allows to click on links in the status text as in the web catalog (left old: no color, non-clickable, right new: color and clickable):
   &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;   

I used the same colors as for the status markers in the search result list view. The green could be a bit less bright to be better readable, like the green color in the web catalog on the left side. Would you want to change it or leave it as is?

### DIFF
--- a/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SLUB.kt
+++ b/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/SLUB.kt
@@ -189,6 +189,7 @@ open class SLUB : OkHttpBaseApi() {
     }
 
     internal fun parseResultById(json: JSONObject): DetailedItem {
+        val colorcodes = mapOf("1" to "#00CC00", "2" to "#FF8900", "3" to "#CC0000")
         val dateFormat = DateTimeFormat.forPattern("dd.MM.yyyy")
         var hasReservableCopies = false
         fun getCopies(copiesArray: JSONArray, df: DateTimeFormatter): List<Copy> =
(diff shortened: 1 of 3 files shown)


StefRe:
On second thought I think might be better to not directly put the color html into the status field in the API but rather supply a per-copy status code (Status.RED etc) in the copy object and do the coloring in the app.

johan12345:
I would also prefer not to pass the color as HTML tags, but introduce an optional separate field with Status.RED/Status.YELLOW/Status.GREEN for each copy instead. This could then also be shown as a colored dot in the copy card (like in search results) instead of coloring the whole text (which, as you said, reduces the readability).

Supporting clickable links in the text looks fine to me, though.

StefRe:
> This could then also be shown as a colored dot in the copy card (like in search results)

I personally find the colored text (left) looks better than the colored status icons (right) but if this is OK for you I'll amend this PR accordingly:

  &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;

raphaelm:
I have no strong preference, I think both look pretty fine. I think if we go with colored text, though, we'd in fact need a darker shade of green.

johan12345:
I think would prefer the colored icons, as they can be directly recognized from the search result list. For colored text we would definitely need to darken both the green and orange colors.

StefRe:
This PR is partly obsolete since SLUB changed the json output in Jan 2021. I'll split it into two separate new PRs: one for copies status colors and one for clickable links in the copy status field.

Issue #543, opened 2018-12-15
Integrate ExLibris Primo via REST API


johan12345:
Basic support for ExLibris Primo (#107) was already implemented using screenscraping.

It seems that Primo also has a REST API that we can use. This becomes especially important as Primo now has a new user interface that our app is not yet compatible with, and some libraries (e.g. TU Berlin) are starting to switch off the old interface.

useful links:
https://developers.exlibrisgroup.com/primo/apis/webservices/rest/pnxs
https://tu-berlin.hosted.exlibrisgroup.com/primo_library/libweb/webservices/rest/v1/configuration/TUB

It seems that VideLibri has already implemented the REST API, so we can look there for usage examples.

StefRe:
I'm much interested in the new Primo interface as it's also used by [HTW Dresden](https://primo.bib.htw-dresden.de/primo-explore/search?vid=49HTW_VU1&lang=de_DE&fromRedirectFilter=true). See also https://primo.bib.htw-dresden.de/primo_library/libweb/webservices/rest/v1/configuration/49HTW_VU1.

From [Exlibris' deverloper web site](https://developers.exlibrisgroup.com/primo/apis/#logging) it seems, however, that you need an account with the Developer Network for each library in order to be able to create an API key:
> If you are a developer at a Primo institution, an account has already been created for your institution. Contact your institution’s technical contact person in order to get an invitation to join the institution group. Once you have joined your institution’s group, you will be able to create and edit your institution’s applications.
<hr>  

> It seems that VideLibri has already implemented the REST API   

At least for HTW Dresden it doesn't work at all (version 2,185): I get an error when trying to log in via VideLibri ("Beim Zugriff auf das Konto ... ist leider ein Fehler aufgetreten:  " and "Die Bibliothek zeigt diese Nachricht auf der Katalogwebseite an:  ", but there's no information after the colons). Searching doesn't throw errors but it yields completely unrelated data, obviously from some kind of demo database. I didn't look further into it but the error message suggests that they don't use an API but instead try to scrape the web site.

StefRe:
I looked into it and it turns out that the new primo interface uses a lot of javascipt to dynamically generate it web sites by calling the `webservices` REST API.  

[...]

PR #608, opened 2020-12-15
TOOLS: Fix loadGeoPossibilities

StefRe:
Since mid 2018 you need an API key to use the Google maps API. If you have such a key,
store it in a file `googlemaps_api_key.txt` in the same directory as `add_library.py`
(this file is not checked into version control).
If you don't have a Google API key, we use OSM Nominatim instead (no key required).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,4 @@ opacclient/opacapp/src/main/assets/bibs/*.json
 opacclient/opacapp/src/main/assets/last_library_config_update.txt
 opacclient/opacapp/src/main/resources/sentry.properties
 
+tools/googlemaps_api_key.txt
--- a/tools/add_library.py
+++ b/tools/add_library.py
@@ -34,7 +34,7 @@ def getInput(required=False, default=None):
     return inp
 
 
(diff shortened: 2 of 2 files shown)


johan12345:
Looks good to me, thanks! We don't really use these tools ourselves anymore, as we have a new one integrated into the web server which serves the configuration files - but it's still good to keep them at least in a usable state.

StefRe:
I just needed the tool as for the new [PrimoExplore API](https://github.com/opacapp/opacclient/issues/543#issuecomment-725585271) I'll have to add quite a number of keys from the Primo config file to the JSON configuration.

> We don't really use these tools ourselves anymore, as we have a new one integrated into the web server which serves the configuration files
Is this new tool publically accessible? If so then the Wiki page [Adding a new library](https://github.com/opacapp/opacclient/wiki/Adding-a-new-library) should be updated as it refers to the python scripts.

johan12345:
No, that is not publically accessible as it saves the configuration files directly to our database. I added some more information to the [wiki page](https://github.com/opacapp/opacclient/wiki/Adding-a-new-library) now.

PR #613, opened 2021-02-28
SLUB: Add references to other SLUB catalog items as volumes

StefRe:
This is a workaround as the only possibility to add links to other detailed
items from a detailed item. It's possible as we can display volumes and
copies at the same time. References are some kind of details which is why
the display order of copies and volumes was swapped: first show volumes
(which may be references) right after the details and only then copies.
Increase volumes text view to maximum two lines high to be able to put
the title of the reference and the reference itself on different lines
for better readability.

Change resolution of bc or rsn and the new libero identifiers to following
head redirects and retrieving the final url to get the id identifier. This
requires one request more than the old method (get location parameter from
non-redirecting head request) but is much shorter to implement.

IDs of references have a single query component (e.g. ?libero_mab21364775)
and no key=value pairs. The `URLEncodedUtils.format` implementation used by
Android has a bug and adds a trailing equal sign (`=`) to the query component
which leads to infinite redirects, see #612 for details. ~~To prevent this
we only use `cleanUrl` if there are any `=` signs in the url, otherwise we use
the raw url in head requests.~~ This is being addressed in commit 91ddcaf.  

Example with two references "Fortsetzung" and "Darin aufgegangen":

### DIFF
--- a/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/OkHttpBaseApi.java
+++ b/opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/OkHttpBaseApi.java
@@ -233,7 +233,7 @@ public abstract class OkHttpBaseApi extends BaseApi {
      */
     public Response httpHead(String url, boolean ignore_errors) throws IOException {
         Request request = new Request.Builder()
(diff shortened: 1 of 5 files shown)


StefRe:
**This PR depends on #614**, so it can't be merged prior to #614.

PR #628, opened 2021-03-31
Make links in copy views clickable

StefRe:
Needed for SLUB where copy status may be empty and a link is provided
to the institute/chair where this book might be available.
(prior to Jan 2021 the link was provided in the status field e.g. as
"Verfügbarkeit bitte in \<a href="...">Professur für ...\</a> erfragen")
See also https://stackoverflow.com/a/20647011/3944322.

See also second item/commit of #606.


StefRe:
@raphaelm you merged `master` into my local feature branch and not the other way round. Was this on purpuse or just a mistake?

PR #640, opened 2021-06-12
Fixes and improvements for Adis at HfM Karlsruhe

johan12345:
[HfM Karlsruhe](https://bsz.ibs-bw.de/aDISWeb/app?service=direct/0/Home/%24DirectLink&sp=SOPAC33) seems to have gotten an update of Adis, which made some adjustments in the API for the account view and search necessary.

Also, they added a free search field that is accessible in the advanced search screen (and can be combined with the advanced fields), so I also implemented support for that.


johan12345:
the Travis CI failure seems to be a missing accepted license...

PR #643, opened 2021-08-31
Add WISE as a new library type

henkkh:
WISE is used by 75% of the public libraries in NL, dutch speaking BE and some US public libraries. https://www.oclc.org/go/nl/wise-nl.html


raphaelm:
Wow, thanks! I'm traveling the next days and a little overcommitted on things generally right now, but I'll try to take a closer look at this asap :)

henkkh:
Hi Raphael,

Your welcome. Take your time (but not ages :-))

Henk

Op di 31 aug. 2021 om 23:02 schreef Raphael Michel ***@***.***

opacclient/opacapp/src/main/res/values-nl/strings_api_errors.xml
```diff
@@ -0,0 +1,89 @@
...
+    <string name="order_category_desc">By category (descending)</string>
+    <string name="interlib_branch">Interlibrary: %s</string>
+    <string name="stacks_branch">Stacks: %s</string>
+    <string name="provision_branch">Provision: %s</string>
+    <string name="unsupported_in_library">This feature is not supported in this library.</string>
+    <string name="reservation_warning">This reservation might cost you a fee.</string>
+    <plurals name="reservations_number">
+        <item quantity="one">%d reservation</item>
+        <item quantity="other">%d reservations</item>
+    </plurals>
```
raphaelm, line 46: I don't think we should add English strings to the Dutch language file :)

opacclient/opacapp/src/main/res/values-nl/strings.xml
```diff
@@ -0,0 +1,358 @@
...
+    <string name="yes">Ja</string>
+    <string name="no">Nee</string>
+    <string name="cancel">Annuleer</string>
+    <string name="delete">Verwijder</string>
+    <string name="accept">OK</string>
+    <string name="doing_cancel">Reservering wordt geannuleerd…</string>
+    <string name="doing_prolong">Verlenging wordt doorgevoerd…</string>
+    <string name="doing_download">Download information is being retrieved…</string>
+    <string name="reader_needed">To read the e-book, you need a separate app, for example \'Aldiko\'. Click \'Ignore\' if you already have such an app, this message won\'t show up again in this case. Click \'Download\' to open up the Play Store to download Aldiko.</string>
+    <string name="reader_needed_overdrive">To open this medium, you need a separate app, for example OverDrive\'s own app. Click \'Ignore\' if you already have such an app, this message won\'t show up again in this case. Click \'Download\' to open up the Play Store to download the OverDrive app.</string>
```
raphaelm, line 47: If you do not want to translate something, just omit it from the file. Including English text in Dutch files makes it very hard to find what translations are missing.

opacclient/libopac/src/test/java/de/geeksfactory/opacclient/apis/WiseTest.kt
```diff
@@ -0,0 +1,61 @@
...
+class WiseTest(private val file: String) : BaseHtmlTest() {
+    val wise =  Mockito.spy(Wise::class.java)
+
+    init {
+        wise.init(Library().apply {
+            data = JSONObject().apply {
+                put("baseurl", "https://catalogus.biblionetdrenthe.nl")
+                put("imgurl", "/cgi-bin/momredir.pl?")
+                put("api-key-id", "4251d316-b964-44cc-b5fc-2f888d3716a8")
+                put("api-key", "218f81af-dcab-4cca-97f3-a4cffe19a0c9")
```
raphaelm, line 28: Our general tests should not do real network calls to avoid breaking tests when library servers are down. I see that you're mocking things, which is good, but in that case we can probably also skip hardcoding API keys?

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
```diff
@@ -0,0 +1,563 @@
...
+
+/**
+ * OpacApi implementation for Wise owned by OCLC https://www.oclc.org/en/wise.html
+ *
+ *
+ * @author Henk Klijn Hesselink, April 2021
+ */
+
+
+class WiseKeyInterceptor : Interceptor {
```
raphaelm, line 40: Hmm, I'm not completely sure if an interceptor is the best design to add an Authorization header. Is this a best practise with okhttp? It feels like it makes it harder to follow the code and debug what is actually being sent.

I'd personally prefer extending OkhttpBaseApi with something like

```
open fun prepareRequestBuilder(): RequestBilder
```

that can then be used to add the header to every request.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
```diff
@@ -0,0 +1,563 @@
...
+    override fun intercept(chain: Interceptor.Chain): Response {
+
+        val originalRequest = chain.request()
+        val requestBuilder = originalRequest.newBuilder()
+                .header("WISE_KEY", genWisekey())
+        if ( auth != null) requestBuilder.addHeader("Authorization", auth!!.token)
```
raphaelm, line 46: Please make sure all new code is auto-formatted with Android Studios default formatting (Code > Reformat file)

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
```diff
@@ -0,0 +1,563 @@
...
+
+    fun genWisekey(): String {
+        val epochDay: String  = floor(Date().time / 8.64e7).toLong().toString()
+        val data = epochDay + applicationName
+
+        val mac = Mac.getInstance("HmacSHA256")
+        val secret_key = SecretKeySpec(apiKey.toByteArray(charset("UTF-8")), "HmacSHA256")
+        mac.init(secret_key)
+        val signature= bytesToHex(mac.doFinal(data.toByteArray(charset("UTF-8"))))
+        return "$apiKeyId:$signature"
```
raphaelm, line 59: Is there any documentation on this algorithm that we could link to here for future reference?

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
```diff
@@ -0,0 +1,563 @@
...
+    private var holdstatuses: Map<String, String>? = null
+
+    override fun init(library: Library, factory: HttpClientFactory, debug: Boolean) {
+        http_interceptor = WiseKeyInterceptor()
+        super.init(library, factory, debug)
+        libraryData = library.data
+        homeBranch = library.libraryId
+        baseurl = libraryData.getString("baseurl")
+        imgurl = libraryData.getString("imgurl")
+        val cats = libraryData.optJSONArray("search-locations")
```
raphaelm, line 111: Is there a specific reason we mix the terms "cat(egorie)s" and "locations" together here?

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
```diff
@@ -0,0 +1,563 @@
...
+import javax.crypto.spec.SecretKeySpec
+import kotlin.experimental.and
+import kotlin.math.floor
+
+data class Authentication(val libraryId: String, val patronSystemId: String,val token: String, val membershipExpiryDate: org.joda.time.LocalDate )
+
+lateinit var apiKeyId: String
+lateinit var apiKey: String
+lateinit var applicationName: String
+var auth: Authentication? =  null
```
raphaelm, line 30: These leak into the variable scope of the whole package, don't they? This seems unclean and prone to cause confusion. Can we make tie them to the Wise class, please? :)

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
```diff
@@ -0,0 +1,563 @@
...
+import org.json.JSONArray
+import org.json.JSONException
+import org.json.JSONObject
+import java.util.*
+import javax.crypto.Mac
+import javax.crypto.spec.SecretKeySpec
+import kotlin.experimental.and
+import kotlin.math.floor
+
+data class Authentication(val libraryId: String, val patronSystemId: String,val token: String, val membershipExpiryDate: org.joda.time.LocalDate )
```
raphaelm, line 25: Better name it WiseAuthentication to avoid confusion

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
```diff
@@ -0,0 +1,563 @@
...
+            return listOf(freeSearchField, ietsTitelOrAuteur, locations, materiaal)
+        }
+        return listOf(freeSearchField, ietsTitelOrAuteur, materiaal)
+
+    }
+
+
+    override fun getSupportFlags(): Int {
+        return OpacApi.SUPPORT_FLAG_ENDLESS_SCROLLING or OpacApi
+                .SUPPORT_FLAG_WARN_RESERVATION_FEES or OpacApi.SUPPORT_FLAG_ACCOUNT_PROLONG_ALL
```
raphaelm, line 316: You should not set SUPPORT_FLAT_ACCOUNT_PROLONG_ALL if you don't actually implement it ;)

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
```diff
@@ -0,0 +1,563 @@
...
+            searchLocations = emptyList()
+            catsList.forEach {
+                json -> val key = json.names()[0];  searchLocations = searchLocations + Pair(key as String, json.get(key) as String)
+            }
+        }
+        if ( libraryData.has("api-key-id") ) {
+            apiKeyId = libraryData.getString("api-key-id")
+            apiKey = libraryData.getString("api-key")
+            applicationName = libraryData.getString("app-name")
+        }
```
raphaelm, line 123: What happens in the `else` case? These are lazyinit properties, do things just crash?

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
```diff
@@ -0,0 +1,563 @@
...
+        val json = JSONObject(response)
+        val objects = List(json.getJSONArray("objects").length(), json.getJSONArray("objects")::getJSONObject)
+        val searchResultList = objects.map { o -> o.getJSONObject("fields") }.filter { fields -> getFieldValue(fields.getJSONObject("medium_srt")) != "Website" } .map { fields -> parseSearchResult(fields) }
+
+        with(json.getJSONObject("paging")) {
+            return SearchRequestResult(
+                    searchResultList,
+                    getInt("total"),
+                    getInt("total") / getInt("perpage"),
+                    getInt("offset") / getInt("perpage")
```
raphaelm, line 237: Is this correct, wouldn't we need to round *up*?

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
```diff
@@ -0,0 +1,563 @@
...
+        return searchGetPage(1)
+    }
+
+    override fun searchGetPage(pagina: Int): SearchRequestResult {
+        val AMOUNT = 10
+        val page = pagina -1
+        val response = httpPost(
+                "$baseurl/cgi-bin/bx.pl",
+                ("prt=INTERNET&var=portal&vestnr=$homeBranch&fmt=json&search_in=$scope&amount=$AMOUNT&catalog=default&event=osearch" +
+                        "&preset=all&offset=${page*AMOUNT}$mediaTypeQS&qs=$queryString&vcgrpf=0&backend=wise&vcgrpt=$vcgrpt").toRequestBody("application/x-www-form-urlencoded".toMediaTypeOrNull()),
```
raphaelm, line 225: I think there's some missing encoding here, what if the search query contains e.g. a ``&``? Better use something like FormBuilder!

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
```diff
@@ -0,0 +1,563 @@
...
+            )
+        }
+    }
+
+
+    override fun cancel(media: String?, account: Account?, useraction: Int, selection: String?): OpacApi.CancelResult {
+        val itemId = media!!.split(":")[2]
+
+        httpDelete("$baseurl/restapi/patron/${auth!!.patronSystemId}/library/${auth!!.libraryId}/hold/$itemId")
+        return OpacApi.CancelResult(OpacApi.MultiStepResult.Status.OK)
```
raphaelm, line 247: We should not return OK without any check if this actually worked, maybe at least check the HTTP response code?

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
```diff
@@ -0,0 +1,563 @@
...
+            if ( s != "") {
+                val newDueDate = org.joda.time.LocalDate.parse(s)
+                if ( newDueDate.isAfter(dueDate)) prolongData = id
+            }
+        } }
+    }
+
+
+    override fun checkAccountData(account: Account) {
+        auth = login(account)
```
raphaelm, line 539: have you checked that this fails properly for wrong credentials?

raphaelm (changes requested):
Thank you again!

I've had a first read of the code. I did not test it and I did not yet read all of it, but I have some comments already that I wanted to share with you. Nothing dramatic, but it's lacking a consist code style and a few details will likely cause problems in the future. Let me know if anything is not clear from the review.

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
henkkh: will do

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
henkkh: yes will do

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
henkkh: will do

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
henkkh: No there is no official public documentation on this algorithm. it can be found though hidden in js files op Wise web OPAC. All official 3rd party vendors which needs to consumes endpoint of the RESTapi do get a copy from OCLC. OPAC  App is not on that list (yet)

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
henkkh: will check

opacclient/opacapp/src/main/res/values-nl/strings_api_errors.xml
henkkh: Oops, will remove those

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
henkkh: These `api` settings items are always needed. I will remove if-check

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
henkkh: integer/integer will result in integer: it will do an implicit `floor`

opacclient/libopac/src/main/java/de/geeksfactory/opacclient/apis/Wise.kt
henkkh: Oops!

PR #649, opened 2022-04-20
Adis: media/reservation list improvements

xduugu:
Several improvements to the Adis media list:
- Support for media type prefixes
- Only extract author by splitting at slash if the title string actually contains a slash (and not another part of the item like the signature)
- When extracting the author by splitting the title at a colon, do not use the whole string as author if the title does not contain a colon


raphaelm:
Hi @xduugu, sorry, I missed that this PR was still around while I was working on a client bug report. My patch https://github.com/opacapp/opacclient/commit/ecb7e6eb177fa142429ed3bca2c84739e02a3c89 should already fix 2/3 of the problems you tackle, could you check if that is the case and reduce the PR to what is left?

xduugu:
Hi @raphaelm, I finally managed to rebase/rewrite my PR. The third commit now also sets the media type of the item according to the prefix even though it is currently not used when displaying the media or reservation list.

